Issue #14, opened 2023-07-28
Cross post from RSS


svandragt:
Subscribe to RSS feeds and create posts from rss items

- [x] subscription ui in config.ini
- [ ] ~setting for how often to check for new items in `config.ini`~ Use cron to manage how often to check, KISS
- [x] rss crawling
- [x] post creation

PR #18, opened 2023-07-31
Cross post from RSS

svandragt:
Fixes #14 

- [x] Fixes #21 Prefix feed name slug to feed item slug
- [x] Fixes #20 It must not be possible to use a slug for a new item that already exists. 
- [ ] last processed setting should be per feed, otherwise when adding additional feeds no existing content is imported.
- [ ] feed items updated date should be from the feed not based on the publishing date
- [ ] Result of the above, manual latest created items must be latest in the feed.
- [ ] slug must include the feed name assigned in the conflig

Build a network by subscribing rss/atom feeds and bring other content into your Lamb instance. I'm happy to get some feedback on this feature.

It's intention is to have a Lamb blog for say a team which can pull in posts from all the team members. Lamb is single user but this lets it support multiple authors. 

You can also use it to pull in your content on other sites.

The display of the posts and the mechanics are a first pass. I'm going to setup a site to collect all my RSS feeds.

> [!IMPORTANT]  
> [Posts table changed name from `bleat` to `post`](https://github.com/svandragt/lamb/commit/6bc6aa4b19390af3124bcf6ebe8a241d9165cd63). Run the following shell commands to migrate:

```shell
cd data
cp lamb.db old.db
sqlite3 lamb.db 'ALTER TABLE `bleat` RENAME TO `post`'
cd ..

# Now update the codebase
git pull
```

PR #86, opened 2024-10-16
Reapplying the flock branch

svandragt:
Reimplementation of #18 on top of `main`

- setup with a network of feeds so that external content is periodically cross-posted to your blog

Current issues:

- [x] Editing posts results in the original content to be reapplied huh
- [x] the whole prepare_bean logic is spaghet
- [x] Still need to test adding feeds and what not.
- [x] double escaping issue with code fragments
- [x] Codeception tests are failing
- [x] Support search filter like each word

This PR is growing in scope, would like to merge it.


svandragt:
Fixes #14

PR #284, opened 2026-06-01
Gate draft/scheduled permalink visibility behind login

svandragt:
## Problem

`respond_status` (`/status/<id>`) only rejected **deleted** posts, so any **anonymous** visitor could view a draft or scheduled post via its numeric permalink. Meanwhile `respond_post` (slug) rejected draft/deleted/scheduled for **everyone** — so the logged-in author couldn't preview their own draft by slug either. The two permalink shapes disagreed.

## Fix

Single `Lamb\is_visible(OODBBean)` predicate — the single-post counterpart to `visible_clause()` (#283):

```
never show deleted
show drafts & future-scheduled only to the logged-in author (permalink preview)
anonymous → 404
```

Used in both `respond_status` and `respond_post`.

## Alignment with docs

`docs/scheduling.md` already says a scheduled post is hidden at "its public URL" and frames `/status/<id>` preview as a logged-in capability. This change makes the code enforce what was already documented; no doc edit needed.

## Micropub impact (tracked for monitoring)

Micropub `createCallback` returns a draft's `/status/<id>` as the `Location` URL. An **anonymous** GET of that URL now 404s. Micropub **update/delete is unaffected** — it locates posts via `findPostByUrl` → `R::load`, independent of `respond_status`. Follow-up issue filed to watch for any client that relied on anonymously GET-previewing a freshly created draft.

## Testing

- New `PostVisibilityTest` (9 cases): predicate + both handlers, anonymous vs logged-in.
- Full suite: 617 passing (+9). PHPStan clean. Lint clean.

### DIFF
--- a/src/lamb.php
+++ b/src/lamb.php
@@ -280,6 +280,30 @@ function is_scheduled(OODBBean $post): bool
 }
 
 /**
+ * Returns true when a post may be shown for a direct permalink request
+ * (/status/<id> or a slug URL).
+ *
+ * Deleted posts are never visible. Drafts and posts scheduled for the future
+ * are visible only to the logged-in author, so they can preview their own
+ * work by permalink; anonymous visitors get a 404. This is the single-post
+ * counterpart to visible_clause() (the SQL allow-list for listings), with the
+ * added logged-in preview exception.
+ *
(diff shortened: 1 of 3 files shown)


svandragt:
Follow-up monitoring issue: #285

PR #343, opened 2026-06-05
Record and surface per-feed crawl status (#119)

### LINKED ISSUE #119: what happens when incompatible feeds are encountered in a cron job, does it leave a message for the admin.

svandragt:
Closes #119.

Fixes the silent-failure bug in feed ingestion and adds owner-visible crawl status, following the phased plan locked in the issue.

## Phase 1 — fix the bug + record status (data layer)

- **Failure detection**: a feed is treated as failed when `!$feed->data || $feed->error()`. Previously the per-feed timestamp advanced unconditionally, so a feed failing for weeks recorded the same fresh timestamp as a healthy one.
- **Decoupled the two meanings** previously crammed into `last_processed_date_*`:
  - **success watermark** (`last_success`) — only advances on a successful crawl; drives "process items newer than X".
  - **last attempt** (`last_attempt`) — advances every run; gates the 30-minute fetch-frequency skip, so a failing feed is retried on schedule instead of looking up to date.
- **Dedicated `feedstatus` bean** (Option B from the issue) keyed by `md5(name . url)`, holding `last_attempt`, `last_success`, `last_error`, `error_message`, `item_count`. Config stays the source of truth for *which* feeds exist; the bean only holds health. A freshly dispensed bean seeds its watermark from any legacy `last_processed_date_<key>` option so existing installs don't re-ingest (and duplicate) every item on the first run after upgrade.
- **Pruning**: status rows for feeds removed from config are cleaned up in the existing `/_cron` maintenance pass (`prune_feed_status()`).

## Phase 2 — surface it (two-tab settings page)

`/settings` is restructured into a **CSS-only two-tab layout** (hidden-radio + `:checked`, no JS — degrades gracefully and keeps the single-CSS-file theme model):

1. **Configuration** — the existing INI editor, unchanged in behaviour.
2. **Logs** — read-only feed crawl status: feed name + URL, last success (`human_time()`), items on last run, and the most recent error with when it occurred.

The tab styles are scoped inline in `parts/settings.php` so the page works under every theme (`base`/`2024`/`2026` all fall back to this part).

## Tests (red-green TDD)

- `tests/Unit/FeedStatusTest.php` — a failed fetch must **not** advance the success watermark and **must** record an error; `last_attempt` advances on both success and failure; successful crawls create/count items and clear prior errors; watermark filtering; legacy-option seeding; `get_feed_statuses()` ordering; `prune_feed_status()`.
- `tests/Unit/ResponseHandlersTest.php` — `respond_settings()` now exposes `feed_statuses`.

Full unit suite green (799 tests, 1227 assertions); `composer lint` and `composer analyse` clean.

## Docs

Documented the Logs tab in `docs/cross-posting.md`.

https://claude.ai/code/session_01A2HSMULact5uuFEudigMBL

---
_Generated by [Claude Code](https://claude.ai/code/session_01A2HSMULact5uuFEudigMBL)_


svandragt:
Want to manually test the ux later today

PR #376, opened 2026-06-06
Add 3-day dependency cooldown (.npmrc + Dependabot)

svandragt:
Implements a supply-chain cooldown so freshly published package versions aren't installed before the ecosystem has had a few days to catch malicious releases.

## Changes

- **`.npmrc`** — `min-release-age=3` (npm 11.10+, days) and `minimum-release-age=4320` (pnpm, minutes). Both keys are set because this repo installs with pnpm 10.x, which uses its own key/unit, while the npm key covers anyone installing with npm.
- **`dependabot.yml`** — `cooldown: default-days: 3` on the **npm** ecosystem (so Dependabot doesn't open PRs for versions the package manager then refuses to install) and on the **composer** ecosystem (covers PHP until Composer ships its own minimum-release-age dependency policy — 2.10 added the policy framework; release-age is the planned follow-up).

Dependabot's cooldown only applies to version updates — security updates still flow through immediately.

## Verification

- `pnpm install --frozen-lockfile` passes locally with the new `.npmrc` in place.
- CI will exercise the same install path.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    cooldown: # Covers PHP until Composer ships its own minimum-release-age policy
+      default-days: 3
     groups:
       minor-and-patch:
         update-types:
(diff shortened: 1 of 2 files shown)


svandragt:
The cooldown applies at resolution time, not to versions already in the lockfile.

PR #422, opened 2026-06-12
Store sessions in a dedicated dir under the data directory

svandragt:
## Summary

Lamb sets `session.gc_maxlifetime` to a week (#419) to keep the author logged in, but left `session.save_path` at the system default — typically **shared** with every other PHP app on the host. PHP's session GC sweeps the entire save_path against the *running request's* `gc_maxlifetime`, so:

- when **Lamb's** GC fires (week-long), it spares other apps' short-lived sessions → they linger longer than intended;
- when **another app's** GC fires (e.g. ~24 min default), it reaps **Lamb's** week-long sessions early → defeating the persistence goal.

This points session storage at a dedicated `<data_dir>/sessions` directory (created `0700`, alongside `lamb.db`). That isolates Lamb's GC to its own files in both directions, and because it lives under the persistent data dir, a deploy or container restart that wipes ephemeral storage no longer logs the author out before the `lamb_logged_in` marker cookie expires.

This keeps the existing **two-cookie** auth model untouched (a stolen marker alone still can't authenticate — it only gates `session_start()`; the real `LAMBSESSID` session is still required), rather than weakening it to a single bearer token.

## Changes

- `bootstrap.php`: new `configure_session_save_path($data_dir)`; `bootstrap_session()` now takes the data dir.
- `index.php`: resolve the data dir once and pass it to both `bootstrap_db()` and `bootstrap_session()`.
- `SessionBootstrapTest.php`: assert the save path is a dedicated dir under the data dir and is created.

## Testing

- `vendor/bin/codecept run Unit` — 988 tests pass
- `composer lint` + `composer analyse` (PHPStan level 8) — clean

### DIFF
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -179,13 +179,40 @@ function start_session(): void
 }
 
 /**
- * Initializes and secures a session, starting it only for (previously) logged-in users.
+ * Points PHP's session storage at a dedicated directory under the app's data dir.
  *
+ * The default save_path is usually shared with every other PHP app on the host.
+ * Because PHP's GC sweeps the whole save_path against the running request's
+ * gc_maxlifetime, Lamb's week-long lifetime would otherwise spare other apps'
+ * short-lived sessions (they linger), and their shorter GC would reap Lamb's
+ * week-long sessions early. A dedicated directory isolates Lamb's GC to its own
+ * files in both directions, and putting it under the persistent data dir means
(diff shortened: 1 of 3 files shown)


svandragt:
Currently it's more secure than restoring login from the cookie (together with a regular length session) but if this is not working to a satisfactory level then that alternative strategy would simplify the session handling at the expense of a layer of security.

PR #478, opened 2026-06-22
Persist logins for a week with HMAC-signed session marker

svandragt:
## Summary

- Give the session cookie and server-side session a one-week lifetime (`REMEMBER_LIFETIME`) — no remember-me checkbox, always on
- Gate `should_start_session()` on an HMAC-signed marker (`<id>.<hmac>`) rather than bare cookie presence; a forged cookie fails cheaply without touching `session_start()`
- Fix a bug where the already-authenticated redirect path skipped issuing the marker, leaving the visitor stuck as anonymous until they cleared cookies

## Test plan

- [ ] Log in and close the browser; reopen — should still be authenticated
- [ ] Restart PHP/the server; existing session should survive
- [ ] Tamper with `LAMBSESSID` cookie value — should be treated as anonymous, no new session file written
- [ ] Change the login password — should log out all existing sessions

### DIFF
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -70,8 +70,14 @@ function configure_session(): void
     ini_set("session.cookie_secure", $secure ? 1 : 0);
     ini_set("session.use_strict_mode", 1);
 
+    // Remember the login for a week: persist the server-side session that long so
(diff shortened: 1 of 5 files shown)


svandragt:
Work already merged via #419 and #420. Closing as redundant.

PR #479, opened 2026-06-22
Standardise on FrankenPHP for dev and self-hosted setups

### LINKED ISSUE #54: Frankenphp support
FrankenPHP is caddy with a PHP server built in, so you don't need to manage php-fpm as you do with caddy itself.

Lamb supports it somewhat already if you run it locally this way: 

```
sudo -E $(which frankenphp) php-server -r src/ -l 0.0.0.0:80 -d lamb.test  
```
`sudo` is required because of the use of a privileged port (80, because the automatic HTTPS cert support cannot resolve a localonly .test domain) 

- [x] How to set the environment variable `LAMB_LOGIN_PASSWORD`? Answer: just set it, it works as long as you add `-E` to `sudo`.
- [ ] Documentation is required
- [x] Should we support this instead of caddy? yep
- [ ] create configuration
- [ ] Experiment packaging Lamb as a binary

svandragt:
## Summary

- Switch dev Docker stack from caddy + php-fpm to a single FrankenPHP container; adds `gd` and `git` so the full test suite runs inside
- Root Caddyfile uses `php_server` (FrankenPHP), no fpm socket; `Caddyfile.release` gains immutable static-asset cache header
- Fix `phpcs.xml` hidden-dir exclusion that silently skipped all files under dotted checkout paths (e.g. `.claude/worktrees/`), letting snake_case test names slip through
- Fix `UpgradeScriptTest` scrubbing inherited `GIT_DIR` env so test fixtures don't commit into the real repo when run from a worktree
- Docs: `frankenphp.md` replaces `caddy.md` as the recommended bare-metal path; `nginx.md` stays for classic setups

Closes #54

## Test plan

- [ ] `composer serve:frankenphp` starts and serves the site at localhost
- [ ] `composer test` passes inside the dev container
- [ ] `composer lint` passes and actually checks files (not a silent no-op)
- [ ] Release image build unchanged


svandragt:
Work already merged via #325 and #326. Closing as redundant.

PR #529, opened 2026-07-25
Security review: fix Micropub auth bypass, front-matter privilege escalation, and stored XSS

svandragt:
A security review of the codebase. Five fixes, one per commit, each independently reviewable and cherry-pickable. All 1291 unit tests pass; `phpcs` is clean.

The two most serious issues are remotely triggerable and need no credentials.

## 1. Micropub authentication bypass via a spoofed `Host` header (critical)

`ROOT_URL` was built from `$_SERVER['HTTP_HOST']`, and `verifyAccessTokenCallback()` used it as the expected IndieAuth identity — a token was accepted when its `me` matched `ROOT_URL`.

Since the `Host` header is client-supplied, an attacker who holds a token the configured token endpoint issued for *their own* identity can send `Host: their-site.example` and have the check compare their `me` against their own host. The default token endpoint is the public `tokens.indieauth.com`, which issues a token for any domain the requester can authenticate as, and both bundled Caddy configs use a host-agnostic `:80` site block, so an arbitrary `Host` reaches the app. That granted unauthenticated remote create/update/delete on posts, plus writes into the web root through `/micropub-media`.

Fixed by adding `Config\canonical_site_url()`: the site URL the author states explicitly, via a new `site_url` setting or the `LAMB_SITE_URL` environment variable. Token verification compares against that and nothing else.

**Behaviour change worth a decision:** token verification now *fails closed* when no canonical URL is configured — with nothing trustworthy to compare against, a token's identity cannot be established. So Micropub stops working until `site_url` is set. The refusal is written to `error_log()` as well as the Micropub log (the latter is silent unless `micropub_debug` is on), and both `docs/micropub.md` and `docs/site-configuration.md` now lead with the setting. If you'd rather not break existing Micropub users on upgrade, the alternative is to keep comparing against `ROOT_URL` when `site_url` is unset — but that leaves the bypass open by default, which is why it isn't what this PR does.

`ROOT_URL` itself still falls back to the request host, so nothing else changes for installs without the setting. That fallback now goes through `Http\request_root_url()`, which validates the `Host` against a strict host charset and returns 400 otherwise — a `Host` carrying quotes, angle brackets or CRLF can no longer be reflected into HTML or a response header.

## 2. Front matter could write any column (high)

`apply_frontmatter()` copied every word-character front-matter key onto the bean, so front matter could set *any* column. Setting `id` made the following store an UPDATE of that row instead of an insert; `deleted` trashed a post.

A Micropub token holding only `create` scope reached this: a create whose content *is* a front-matter block is parsed as front matter, so `id:` overwrote an arbitrary post and `deleted:` destroyed one — bypassing the scope checks that correctly refuse `update` and `delete`. Unknown keys also became new columns via RedBean's fluid mode. Front matter now writes only the fields it owns.

`build_matter()` was a second route to the same sink: it assembled the block as `"$key: $value"` with no escaping, and Micropub property values reach it, so a `name` containing a newline injected further front-matter keys. It now emits YAML via `Yaml::dump()`, as the WordPress importer already does. That also fixes the mirror-image bug where a title containing a colon produced invalid YAML, so the block failed to parse and the title was silently dropped.

## 3. Stored XSS through hashtag linkification (high)

`parse_tags()` injected `<a href="…">` into already-rendered HTML wherever a hashtag appeared — including inside attributes Parsedown builds from user text, such as an image `alt` or a link `title`. The injected quote closed the attribute, and the rest of the value was then parsed as attributes of the enclosing element, so `#x/onerror=alert...` inside an `alt` became a real event handler that fires when the image fails to load.

Ingested feed content flows through this function and the result is stored, so a remote feed could plant the payload for every visitor; with the default `feeds_draft` it lands first in the author's own logged-in session. Substitution is now confined to text position, the tag name is escaped, and the tag pattern no longer admits quotes, angle brackets, backticks, `=` or slashes. Emoji and non-Latin hashtags still work.

## 4. Outbound HTTP was unbounded and partly unguarded

- **No response-size caps.** `POST /webmention` and `/_cron` are both unauthenticated and make the server fetch chosen URLs, yet neither the source/target/endpoint fetches nor the JSON-feed crawl passed `max_bytes` — a destination streaming an endless body grew the worker's memory until it fatalled. The fetcher already supported the cap; these call sites never used it.
- **The Micropub token introspection followed redirects.** PHP's stream wrapper re-sends the context's `header` option to the redirect target, across a change of authority included, so a redirect from the token endpoint handed the author's bearer token to whatever host it named. It now refuses redirects and caps the body.
- **WebSub hub pings skipped every URL check.** `hub_urls()` only split on commas and `send_ping()` used the unguarded `post_form()`, so a hub value could name any scheme PHP has a wrapper for and reach an internal address. This was the last outbound sink bypassing the SSRF guard the feed and webmention paths already use.

## 5. Upload directory and web-root hardening

`src/assets` is inside the web root and is the only path the app writes to at runtime, so the extension allowlist was carrying the whole separation between "stored file" and "executed code".

- **Content is checked, not just the filename.** `safe_upload_extension()` inspects only the client-supplied name, so any bytes could be stored under an image extension and served from this origin — an HTML or SVG payload at a URL on your own domain. Every write path now sniffs the content and refuses a mismatch. Container formats still accept an unidentifiable stream, since libmagic cannot always name ISOBMFF or Matroska; scripted payloads never sniff that way.
- **PHP will not execute from the upload tree.** Both server configs would have run a `.php` file found under `/assets`; each now returns 404 for one.
- **Security response headers.** `X-Content-Type-Options`, `Referrer-Policy` and `X-Frame-Options` were set nowhere. `nosniff` matters most here, since it stops a browser second-guessing an upload's `Content-Type`. No CSP is added — the themes emit inline styles, so that needs its own change.
- Smaller items: `/upload` sets a JSON `Content-Type` (it echoes the client filename and defaulted to `text/html`, so the name was parsed as markup) and authenticates before inspecting the request; created directories drop the world-writable bit `0777` leaves under a permissive umask — the data directory holds `lamb.db` and the session files; the release image turns off `expose_php`, which PHP's `php.ini-production` leaves on.

## 6. Escaping fixes

`title_link()` and `date_created()` interpolated the permalink and slug straight into an `href` and a `title` attribute. A slug is stored close to verbatim, so it needs escaping. Only the author or an authorised Micropub client can set one, so this is self-inflicted rather than remotely triggerable, but they were the last unguarded URL sinks in the theme layer.

The Atom feed's `<id>` used `addChild()`, which — unlike `addAttribute()` — does not escape, so a permalink containing `&` emitted an empty `<id/>` and left the feed malformed for every reader. The two `addAttribute()` calls beside it had the opposite bug and pre-escaped a value that gets escaped anyway.

## Found but deliberately not changed

Each of these is a judgement call that belongs to you rather than to a security patch:

- **The release image runs FrankenPHP, and therefore all PHP, as root.** `.docker/Dockerfile.release` has no `USER` and no privilege drop, while the dev image does exactly that (`setcap` + `entrypoint.sh`). Any RCE becomes container root. I did not change it: there is no Docker daemon in this environment to build and test against, and a naive `USER www-data` would break existing deployments whose named volumes and bind mounts are still root-owned. The safe shape is probably a release entrypoint that chowns `/app/data`, `/app/src/assets`, `/data` and `/config` and then drops privileges, mirroring the dev one.
- **`GET /micropub?q=config` is served without a token**, deliberately per the comment on the `handleRequest` override. It discloses the media-endpoint URL and every configured `[syndicate_to]` target. W3C Micropub §3.7 wants a token here, but tightening it may break real clients.
- **`MAX_UPLOAD_PIXELS` (40 MP) exceeds what the release image's 128 MB `memory_limit` can decode** — a 40 MP truecolor buffer is roughly 160 MB, so the pre-decode guard admits images that reliably fatal the request. Lowering it caps the maximum photo resolution you support, which is a product decision.
- **Type confusion in Micropub input causes uncaught `TypeError`s → HTTP 500** for a few well-formed-JSON shapes (an array where `properties.name`, `properties.published`, or `replace.content` expects a string). Robustness rather than a boundary; guards returning `invalid_request` would fix it.
- **Third-party GitHub Actions float on mutable tags**, including the `docker/*` actions in the job that publishes the image with `packages: write`. Pinning to SHAs would close that; Dependabot already watches `github-actions`.
- **`/upload` and `/checkbox` have no CSRF check** — documented as relying on `SameSite=Strict`, which does hold. Adding `require_csrf()` would actually break the flow, since the token is single-use and `/upload` runs mid-composition.
- **A DNS-rebinding TOCTOU remains in the SSRF guard**: `is_public_http_url()` resolves the host, then the fetch resolves again independently. Closing it means connecting to the validated IP with an explicit `Host` header.

Clean on review: SQL injection (every user-controlled value is bound or integer-cast across all 93 query sites), XXE in feed and WXR parsing, TLS verification, committed secrets, password hashing, and the WordPress importer's reachability (CLI-only).

---
_Generated by [Claude Code](https://claude.ai/code/session_01X2iwbhM8UYXqU52HERiKLp)_

### DIFF
--- a/.docker/Caddyfile
+++ b/.docker/Caddyfile
@@ -5,6 +5,16 @@
 :80 {
 	root * /srv/app/src
 
+	header {
+		X-Content-Type-Options nosniff
+		Referrer-Policy strict-origin-when-cross-origin
+		X-Frame-Options SAMEORIGIN
+	}
+
+	# Never execute PHP from the runtime upload directory. See Caddyfile.release.
+	@assets_php path_regexp ^/assets/.*\.(?i:php|phar|phtml)$
+	respond @assets_php 404
+
 	# Static assets are content-hash cache-busted (?ver=) or stored under
 	# hashed names, so they are safe to cache for a year and immutable.
 	@static path /themes/* /scripts/* /assets/*
--- a/.docker/Caddyfile.release
+++ b/.docker/Caddyfile.release
@@ -6,6 +6,21 @@
 	root * /app/src
 	encode zstd gzip
 
+	header {
+		# Uploads are served from this origin, so a browser must never
+		# second-guess their Content-Type and treat one as HTML or a script.
+		X-Content-Type-Options nosniff
+		Referrer-Policy strict-origin-when-cross-origin
+		X-Frame-Options SAMEORIGIN
+	}
+
+	# /assets is the upload directory: the only web-root path the app writes to
+	# at runtime. The upload allowlist already refuses anything but images and
+	# video, but nothing should execute PHP from there even if a file slips
+	# through — php_server would otherwise happily run it.
+	@assets_php path_regexp ^/assets/.*\.(?i:php|phar|phtml)$
+	respond @assets_php 404
+
 	# Static assets are content-hash cache-busted (?ver=) or stored under
 	# hashed names, so they are safe to cache for a year and immutable.
 	@static path /themes/* /scripts/* /assets/*
--- a/.docker/Dockerfile.release
+++ b/.docker/Dockerfile.release
@@ -21,7 +21,9 @@ RUN cp "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
 # Images are converted to WebP and resized server-side, so large photo
 # originals are fine (phone photos easily exceed PHP's 2M default); video is
 # stored as-is and can be much larger still.
-RUN printf 'upload_max_filesize = 100M\npost_max_size = 100M\n' > "$PHP_INI_DIR/conf.d/uploads.ini"
+# expose_php is On in PHP's shipped php.ini-production, which advertises the
+# exact PHP version in X-Powered-By on every response.
+RUN printf 'upload_max_filesize = 100M\npost_max_size = 100M\nexpose_php = Off\n' > "$PHP_INI_DIR/conf.d/uploads.ini"
 
 COPY .docker/Caddyfile.release /etc/frankenphp/Caddyfile
 
--- a/.nginx/snippets/lamb.conf
+++ b/.nginx/snippets/lamb.conf
@@ -3,6 +3,14 @@
 # is stored as-is and can be much larger still.
 client_max_body_size 100m;
 
+# Uploads are served from this origin, so a browser must never second-guess
+# their Content-Type and treat one as HTML or a script. add_header does not
+# merge across levels — a location that sets its own headers drops these — so
+# they are repeated in the static-asset location below.
+add_header X-Content-Type-Options nosniff always;
+add_header Referrer-Policy strict-origin-when-cross-origin always;
+add_header X-Frame-Options SAMEORIGIN always;
+
 location / {
      try_files $uri $uri/ /index.php$is_args$args;
 }
@@ -12,4 +20,7 @@ location / {
 location ~ ^/(themes|scripts|assets)/ {
      expires 1y;
      add_header Cache-Control "public, immutable";
+     add_header X-Content-Type-Options nosniff always;
+     add_header Referrer-Policy strict-origin-when-cross-origin always;
+     add_header X-Frame-Options SAMEORIGIN always;
 }
--- a/.nginx/snippets/php-82.conf
+++ b/.nginx/snippets/php-82.conf
@@ -1,5 +1,14 @@
 index index.php index.html;
 
+# /assets is the upload directory: the only web-root path the app writes to at
+# runtime. The upload allowlist already refuses anything but images and video,
+# but nothing should execute PHP from there even if a file slips through. This
+# must stay ahead of the `\.php$` location below — nginx takes the first regex
+# location that matches.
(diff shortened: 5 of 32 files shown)


svandragt:
Filed the "Found but deliberately not changed" section as issues so it can be followed up independently of this PR:

| Issue | Finding |
|---|---|
| #530 | Release Docker image runs FrankenPHP, and all PHP, as root |
| #531 | Third-party GitHub Actions float on mutable tags, incl. the image-publishing job |
| #532 | `MAX_UPLOAD_PIXELS` admits images the release image's `memory_limit` cannot decode |
| #533 | Micropub input type confusion returns 500 instead of `invalid_request` |
| #534 | `q=config` served without a token, disclosing syndication targets |
| #535 | SSRF guard is DNS-TOCTOU — the validated IP is not the one connected to |
| #536 | `/upload` and `/checkbox` rest entirely on `SameSite=Strict` for CSRF |

#530 is the one I'd prioritise. #536 has no live vulnerability — it is filed to pin the property the design depends on, since adding `require_csrf()` there would actually break the flow.

---

PR #558, opened 2026-07-29
Import a Lamb export (lamb-export/1)

### LINKED ISSUE #554: Import a Lamb export (lamb-export/1) — backup restore and lamb→lamb migration
Follow-up to #440 / #552, which adds `/export` and specifies the `lamb-export/1` format (`docs/export.md`). That gives us a documented way *out*; this issue is the way back *in*.

## Why

Right now a Lamb export is a fine archive but a dead end — restoring it means hand-editing SQLite. The missing piece is backup/**restore** that isn't "copy `data/lamb.db` and hope the RedBean schema still matches", plus lamb→lamb migration between installs.

This also closes the loop the format was chosen for: `posts/**/*.md` is byte-identical to what the WordPress and Known importers already produce via `Import\build_post_body()`, so an importer for it is the exact inverse of the exporter rather than a new parsing path.

## Proposal

A CLI importer matching the two existing ones:

```bash
php import-lamb.php path/to/lamb-export-2026-07-26.zip [--dry-run] [--replace]
```

- **`--dry-run`** — parse and report, write nothing. Same contract as the other importers.
- **`--replace`** — update posts that already exist instead of skipping them.
- Accept an unzipped directory as well as a `.zip`; both are cheap to support and a partially-recovered backup is often already unpacked.

Suggested layout, following the existing split: reader/validation logic in `src/` under its own namespace (`Lamb\Restore` in `src/restore.php` reads well and avoids a `Lamb\Lamb`), with `import-lamb.php` as the thin CLI wrapper — same shape as `src/wordpress.php` + `import-wordpress.php`.

### What it does

1. Read and validate `manifest.json`; refuse a `format` outside `lamb-export/1`, ignore unknown fields (the format is additive within a major).
2. For each manifest entry, read its `.md` and run it through the normal pipeline — `Post\populate_bean()` → `Post\finalize_and_store_post()` — so imported posts go through the same code as any other create.
3. Apply the manifest state *after* `populate_bean()`, which is the whole point of having a manifest: `created`, `updated`, `draft`, `deleted`/`deleted_at`, `feed_name`, `feeditem_uuid`, `source_url`, and the authoritative `slug` (the `.md` may only carry a title, while the manifest has the slug as actually served — including any `-<id>` suffix — so URLs survive the round trip).
4. Restore referenced assets into `src/assets/YYYY/MM/`, without overwriting files already there.
5. Print the same per-item progress plus `created=… existed=… skipped=…` summary the other importers emit, with a skip-reason breakdown.

Like the existing importers, it should **not** emit outbound webmentions or WebSub pings — restored posts are pre-existing content, not new publications.

## Things that need deciding

**Dedupe key.** Slug is the obvious choice (it is the URL identity, and the export pins it), but it does not cover every post: status posts can have an empty slug and fall back to `/status/<id>` in `Lamb\permalink()`, and `finalize_slug()` returns early on `''`. So slug-dedupe needs a companion rule for slugless posts.

Reusing `feeditem_uuid` with a `lamb-` prefix — the convention `Import\import_uuid()` already establishes — is tempting but **has a side effect**: `lock_if_feed_sourced()` (`src/response/posts.php`) sets `feed_locked` on any post with a non-empty `feeditem_uuid` when it is edited. Synthesising one for a locally authored post would quietly change its behaviour. Either a separate column or a different rule for slugless posts is probably wanted.

**Id collisions.** Manifest `id`s come from the source install and will collide with local ones on a non-empty database. They should be treated as provenance, not as keys to restore.

**Merge semantics.** Into an empty install this is a restore and everything is unambiguous. Into a populated install it is a merge, and `--replace` needs a defined meaning: does it overwrite the body only, or the draft/deleted state too? Worth being explicit in the docs.

## Safety

The archive is untrusted input — it may not be one this install produced.

- Reject entry paths that don't match the expected shape (`posts/YYYY/MM/<name>.md`, `assets/YYYY/MM/<file>`) rather than trusting `manifest.json`'s `path`, and never write outside `src/assets/` — classic zip-slip.
- Only restore assets with an allowed upload extension, reusing `Response\safe_upload_extension()`; an export from elsewhere must not be able to drop a `.php` file under the web root.
- Consider a sanity cap on entry count/uncompressed size (zip bomb) since this runs from the CLI on operator-supplied files.

## Testing

Red-green, mirroring `tests/Unit/ExportTest.php`: build an archive with `Export\build_export_archive()`, import it into a scratch DB, and assert the posts come back with the same bodies, slugs, dates and draft/trash state — an export→import round trip as a single test. Plus the rejection cases: bad `format`, traversal entry paths, a `.php` asset, a manifest pointing at a missing file.

## Related

- #440 — the export discussion and format decision
- #552 — adds `/export`, `docs/export.md` and the `DECISIONS.md` entry this builds on

svandragt:
Closes #554. The inverse of `/export` (#552): a CLI importer that reads a `lamb-export/1` archive back in, for backup restore and lamb→lamb migration.

```bash
php import-lamb.php path/to/lamb-export-2026-07-26.zip [--dry-run] [--replace] [--site-url=<url>]
```

An unzipped directory works as well as a `.zip` — a partially recovered backup is often already unpacked. Like the WordPress and Known importers, it emits no webmentions and no WebSub pings: restored posts are pre-existing content, not new publications.

Draft because it has not been exercised against a real archive from a real install yet — only the round-trip tests. See **Before merging** below.

## The two decisions the issue left open

**Dedupe key: a new `import_uuid` column**, not the slug and not `feeditem_uuid`. Reusing `feeditem_uuid` was rejected because `lock_if_feed_sourced()` (`src/response/posts.php:329`) sets `feed_locked` on any post with a non-empty `feeditem_uuid` when it is edited — synthesising one for a locally authored post would silently change its behaviour. The column is added through the existing `Bootstrap\ensure_post_columns()` idiom, so no new migration machinery.

Identity is `md5('lamb-' . origin . '#' . source_id)`. The origin comes from the manifest's `site.url`, which the exporter already records — so no format change, `lamb-export/1` is untouched. `--site-url=` overrides it, because `ROOT_URL` is host-derived and the same install reached on two hostnames would otherwise yield two origins.

**`--replace` overwrites everything the manifest describes** — body, slug, created/updated, draft, deleted/deleted_at, feed provenance. A restore that leaves a post published when the backup says it was trashed is not a restore. Local edits since the backup are lost; that is what replace means, and the docs say so plainly.

## Untrusted input

The archive may not be one this install produced.

**The archive is never extracted.** Entries are read in place through a reader closure, so there is no temp directory and no zip-slip surface to defend — `ZipArchive::extractTo()` is never called and every destination path is rebuilt from validated `YYYY/MM/<name>` components rather than from an archive entry name. The manifest's `path` is not trusted either: it is cross-checked against what the archive actually contains, and an entry that is not there is reported as a missing file rather than restored as an empty post.

Assets are gated twice — `Response\safe_upload_extension()` on the name and a content sniff on the bytes — so a `.php`, or an HTML-sniffing `.jpg`, cannot land under the web root. Existing files are never overwritten. Entry-count and uncompressed-size caps bound zip bombs, checked against zip metadata before anything is read.

## Review found a real bug

An adversarial review of this branch caught a manifest-controlled dedupe bypass. `manifest_items()` built each item with `$post + [...]`, and PHP's array union keeps the **left** operand's keys — so a manifest supplying its own `import_uuid` kept it, bypassing origin namespacing entirely. With `--replace`, a hostile archive could overwrite a post restored earlier from a different site. A supplied `body` key was likewise honoured without the archive ever being read. Fixed by computing last (`array_merge`), with tests for both.

Three others fixed in the same commit:

- `--dry-run` skipped the content sniff, so an HTML-bodied `.jpg` reported `restored=1 rejected=0` on the dry run and `restored=0 rejected=1` for real. Dry-run is the documented "use this first" mode, so its tally has to match.
- Two archives with no `site.url` shared an empty origin, so restoring the second silently reported `existed=50 created=0` and dropped all of it. Empty origins are now salted per-archive from the manifest, which keeps re-runs idempotent without needing a flag.
- `run_import()`'s replace branch is gated on a lookup actually being supplied. Left ungated it would not have thrown — PHP accepts extra arguments to userland functions — it would have silently created a duplicate row and reported it as `replaced`.

## Shared code

`Import\run_import()` gained two optional parameters (an injectable existing-row lookup, and a replace flag) rather than `src/restore.php` owning a private loop, so all three importers keep byte-identical progress and summary output. Both existing importers pass five arguments and are unaffected; there is a regression test pinning that.

## Testing

`tests/Unit/LambRestoreTest.php`, 52 new tests. The headline is a genuine round trip: build posts (published, draft, trashed with `deleted_at`, slugless status, one with an asset), run them through `Export\build_export_archive()`, nuke the database, import, and assert bodies, slugs, dates and draft/trash state all survive.

Plus: re-import is idempotent; `--replace` overwrites a local edit and re-trashes an untrashed post; two archives with different `site.url` and the same post id produce two rows; a forged `import_uuid` and a forged `body` are both ignored; traversal, absolute, encoded and null-byte entry paths are refused; a `.php` asset and an HTML-sniffing `.jpg` are rejected with nothing written; caps fire; a directory source matches the zip.

One test exists specifically to pin a conclusion the design relies on: a local post is pre-created with a colliding slug, and the restored post is asserted to get `hello-world-<id>` **while still keeping the manifest's `created`/`updated`**. Without it, the "no second store needed" argument had no safety net, because every other fixture pins a slug that never triggers a body rewrite.

`composer lint` clean, `composer analyse` (PHPStan level 8) clean with no baseline additions, 1574 unit tests pass (1522 before), coverage 82.68%.

## Known limitation

A post that had no slug gets a new `/status/<new id>` URL, because source ids cannot be restored into a populated database. Only the URL is affected — dedupe is on the manifest's source id, so it still imports exactly once however often you re-run. Documented in `docs/lamb-import.md`.

## Before merging

- [ ] Run against a real export from a real install, not just the test fixtures
- [ ] Try a merge into a populated database, not only a clean restore
- [ ] Confirm the docs page reads right once Pages rebuilds (it builds from `release`, so this will not be live until the next release)

https://claude.ai/code/session_01KMVaMWvgozaAViNV7iXiWG

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,7 @@ lamb/
 │   ├── routes.php        # register_route() / call_route() helpers
 │   ├── lamb.php          # Core helpers: parse_bean, parse_tags, permalink, visibility clauses, redirects
 │   ├── post.php          # Post helpers: populate_bean, parse_matter, slugify, finalize_slug
+│   ├── restore.php       # Lamb export importer: archive reader, manifest validation, post/asset restore
 │   ├── response.php      # Response helpers: pagination, conditional GET/304, 404, upgrade_posts
 │   ├── response/         # Route handlers (respond_*, redirect_*), split by area
 │   │   ├── auth.php      # Login/logout/settings
@@ -109,6 +110,7 @@ lamb/
 ├── composer.json
 ├── phpcs.xml             # Coding standard config
 ├── codeception.yml       # Test runner config
+├── import-lamb.php       # CLI: restore a Lamb export archive (src/restore.php) into the DB
 └── make-password.php     # CLI utility: hash password → .env
 ```
 
@@ -143,6 +145,7 @@ Each file declares a namespace; functions are called with the namespace prefix:
 | `network.php` + `network/*.php` | `Lamb\Network` |
 | `post.php` | `Lamb\Post` |
 | `response.php` + `response/*.php` | `Lamb\Response` |
+| `restore.php` | `Lamb\Restore` |
 | `routes.php` | `Lamb\Route` |
 | `security.php` | `Lamb\Security` |
 | `theme.php` + `theme/*.php` | `Lamb\Theme` |
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -26,12 +26,14 @@ WXR was rejected: it is WordPress's schema, has no clean home for Lamb's draft/d
 ## 2026-07-29 — Lamb export import identity via a separate `import_uuid` column
 
 **Status:** Accepted
-**Context:** Issue #554 asked for a lamb-to-lamb importer that restores a `/export` archive back into the database, idempotently. The two existing importers (WordPress, Known) dedupe re-runs via `feeditem_uuid`, and reusing that column for restored posts too was the obvious first idea — one less column.
 
-It was rejected. `lock_if_feed_sourced()` (`src/response/posts.php:329`) sets `feed_locked` on any post with a non-empty `feeditem_uuid` when it is edited, to stop hand-editing content a feed will overwrite on the next cron run. A restored local post is not feed-sourced — nothing will overwrite it — so giving it a `feeditem_uuid` to get free dedup would also silently feed-lock it, changing edit behaviour for a post that never subscribed to anything.
+**Context:** Issue #554 asked for a Lamb-to-Lamb importer that restores a `/export` archive back into the database, idempotently. The two existing importers, for WordPress and Known, dedupe re-runs through `feeditem_uuid`, and reusing that column for restored posts was the obvious first idea — one less column.
 
-**Decision:** Add a dedicated `import_uuid` column (`Bootstrap\ensure_post_columns()`, `src/bootstrap.php`), computed as `md5('lamb-' . $origin . '#' . $source_id)`. `$origin` is the exporting site's URL (from the manifest's `site.url`, or `--site-url` when given) rather than `ROOT_URL`, because `ROOT_URL` is host-derived and the same install reached on two hostnames would otherwise mint two origins for one site's posts.
-**Consequences:** Restored posts carry both an empty `feeditem_uuid` and a populated `import_uuid`, so they are editable without tripping feed-lock. `run_import()` (`src/import.php`) gained an optional `$find_existing` callable so its shared counters/summary machinery can dedupe on either column without duplicating the loop.
+We rejected it. `lock_if_feed_sourced()` (`src/response/posts.php:329`) sets `feed_locked` on any post with a non-empty `feeditem_uuid` when you edit it, to stop you hand-editing content a feed will overwrite on the next cron run. A restored local post isn't feed-sourced, and nothing will overwrite it, so giving it a `feeditem_uuid` for free dedupe would also silently feed-lock it. That changes edit behaviour for a post that never subscribed to anything.
+
+**Decision:** Add a dedicated `import_uuid` column through `Bootstrap\ensure_post_columns()` in `src/bootstrap.php`, computed as `md5('lamb-' . $origin . '#' . $source_id)`. `$origin` is the exporting site's URL, taken from the manifest's `site.url` or from `--site-url` when given, rather than `ROOT_URL`. `ROOT_URL` is host-derived, so the same install reached on two hostnames would otherwise mint two origins for one site's posts.
+
+**Consequences:** Restored posts carry an empty `feeditem_uuid` and a populated `import_uuid`, so you can edit them without tripping feed-lock. `run_import()` in `src/import.php` gained an optional `$find_existing` callable, so its shared counter and summary machinery can dedupe on either column without duplicating the loop.
 
 ---
 
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -23,6 +23,18 @@ WXR was rejected: it is WordPress's schema, has no clean home for Lamb's draft/d
 
 ---
 
+## 2026-07-29 — Lamb export import identity via a separate `import_uuid` column
+
+**Status:** Accepted
+**Context:** Issue #554 asked for a lamb-to-lamb importer that restores a `/export` archive back into the database, idempotently. The two existing importers (WordPress, Known) dedupe re-runs via `feeditem_uuid`, and reusing that column for restored posts too was the obvious first idea — one less column.
+
+It was rejected. `lock_if_feed_sourced()` (`src/response/posts.php:329`) sets `feed_locked` on any post with a non-empty `feeditem_uuid` when it is edited, to stop hand-editing content a feed will overwrite on the next cron run. A restored local post is not feed-sourced — nothing will overwrite it — so giving it a `feeditem_uuid` to get free dedup would also silently feed-lock it, changing edit behaviour for a post that never subscribed to anything.
+
+**Decision:** Add a dedicated `import_uuid` column (`Bootstrap\ensure_post_columns()`, `src/bootstrap.php`), computed as `md5('lamb-' . $origin . '#' . $source_id)`. `$origin` is the exporting site's URL (from the manifest's `site.url`, or `--site-url` when given) rather than `ROOT_URL`, because `ROOT_URL` is host-derived and the same install reached on two hostnames would otherwise mint two origins for one site's posts.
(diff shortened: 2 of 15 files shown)


svandragt:
## How to test this manually

Nothing here has been run against a real archive yet — only the round-trip tests — so this is the gap that keeps the PR in draft.

**Point it at a scratch database first.** The importer writes directly to the live database with no review queue, so do the first pass against a copy:

```bash
cp -r data /tmp/lamb-test-data
```

and run with that as the data dir rather than your real one.

### 1. Get a real archive

From a running site, log in and hit `/export`, or use the Settings tab. That gives you `lamb-export-<date>.zip`.

### 2. Dry run first

```bash
devbox run -- php import-lamb.php lamb-export-2026-07-29.zip --dry-run
```

The dry-run tally should **match the real run exactly** — that was a bug in the first pass of this branch (the dry run skipped the asset content sniff, so an HTML-bodied `.jpg` reported `restored` on the rehearsal and `rejected` for real). If the two disagree, that fix has regressed.

### 3. Clean restore

Into an empty database. Check that slugs, dates, and draft/trash state all survive, and that referenced images land under `src/assets/YYYY/MM/`.

### 4. Merge into a populated database

**This is the least-exercised path.** Import into a database that already has posts, including one whose slug collides with something in the archive. Expect the restored post to become `<slug>-<id>` while keeping the archive's `created`/`updated` — there is a unit test for exactly this, but it has never been seen against real data.

### 5. Re-run it

Same archive, same database. Should report `created=0 existed=N` and change nothing.

### 6. `--replace`

Edit a post locally, then re-import with `--replace`. The local edit should be gone and the archive's state restored, including draft/trash flags.

### Worth poking at

- **Two archives from different sites.** They should stay separate. If either lacks a `site.url` in its manifest you will see a warning naming `--site-url=` — restoring two site-url-less archives without that flag is the case that used to silently swallow the second one.
- **A slugless status post.** It comes back at a new `/status/<id>` — source ids are not restorable. Only the URL changes; re-importing still will not duplicate it.
- **An archive that is not from Lamb.** A wrong `format` should be refused outright.

An unzipped directory works anywhere a `.zip` does.

svandragt:
## Merge this before the docs style pass

Branch `claude/docs-google-style-guide-uhr9jn` rewrites all of `docs/` plus the root docs to the [Google developer documentation style guide](https://developers.google.com/style). It overlaps this PR on six files: `README.md`, `DECISIONS.md`, `docs/index.md`, `docs/export.md`, `docs/known-import.md`, `docs/wordpress-import.md`.

**Merge this one first.** The changes here are semantic — a new column, a new script, new manifest semantics — while the style branch is almost entirely prose. Re-applying a prose pass over merged semantic changes is straightforward; re-applying semantic changes over rewritten prose is not.

## Its docs are already restyled

Rather than leave that as a follow-up, 091f9c8 on this branch brings the pages here into the new style, so the two branches meet consistently.

- `docs/lamb-import.md` — imperative task headings (`Restore a Lamb export`, `Run the importer`), passive constructions given Lamb as their actor, long em-dash sentences split, contractions, and an opening sentence matching the wording the WordPress and Known pages use for the same offline/re-runnable claim.
- The Related-link additions in `export.md`, `known-import.md` and `wordpress-import.md` use the `link: sentence` form the rest of `docs/` now uses.
- The new `DECISIONS.md` entry gets the same prose treatment, with its layout left matching its neighbours here since the style branch reformats that file wholesale.

No technical content changed — every flag, path, uuid formula and caveat is as written. Docs build clean at 35 pages with no broken links or anchors.

The style is defined in `DOCS_STYLE.md` (6170b85) and applied across `docs/` in ee6bbca, 92c3092 and 09e5564.

## One thing to fix on merge

The style branch adds `nav_order` to every page, numbered 1–36, because the sidebar was previously alphabetical. I've given `lamb-import.md` `nav_order: 31`, which slots it between Export and the WordPress importer. When the branches meet, the pages after it need nudging by one:

| Page | Now | After |
|---|---:|---:|
| `wordpress-import.md` | 31 | 32 |
| `known-import.md` | 32 | 33 |
| `themes.md` | 33 | 34 |
| `theme-functions.md` | 34 | 35 |
| `troubleshooting.md` | 35 | 36 |
| `project-goals.md` | 36 | 37 |

## Related, not blocking

The style branch adds `docs/backup-restore.md`, deliberately scoped away from this PR: it covers the file-level path (`data/lamb.db` and `src/assets/`) and states that an export is a content archive, since site configuration, received webmentions and stored redirects live only in the database. Once this merges, `import-lamb.php` is the natural `## Related` link to add there.

Your **Before merging** checklist still stands — none of the above exercises the importer against a real archive.

---

PR #559, opened 2026-07-29
Stop modelling imported posts as feed items

svandragt:
Stacked on #558 — **base is `feature/554-import-lamb-export`, not `main`**. Merge #558 first; GitHub will retarget this at `main` automatically.

## The bug

Both CLI importers stamp migrated posts as if they were feed items: `feed_name = 'wordpress'` (`src/wordpress.php:235`) / `'known'` (`src/known.php:350`), plus a `feeditem_uuid`. Neither sets `source_url`. Three consequences, none of them intended and none documented:

1. **Every imported post publicly reads "Via wordpress" or "Via known".** `Theme\link_source()` (`src/theme.php:339`) renders an attribution line for any post with a `feed_name`; with `source_url` unset it falls to the plain-text branch at `:354`, so it is not even a link to the original — just the bare word.
2. **They can never send webmentions or WebSub pings.** `src/webmention.php:431` and `src/websub.php:85` bail on a non-empty `feed_name`, and `src/websub.php:157` excludes those rows from the ping query. This is permanent: edit a migrated post years later and it still notifies nobody.
3. **`lock_if_feed_sourced()` (`src/response/posts.php:329`) sets `feed_locked` on first edit.** Inert today, since no configured feed produces a matching uuid — but it is exactly the trap #554 declined to walk into.

Suppressing outbound notifications *during* an import is right. Making it a permanent property of the post is not: migrated posts are your own content, which is the entire point of importing them.

This surfaced from #554's analysis of why the Lamb importer must not reuse `feeditem_uuid` as its dedupe key. The same reasoning applies to the two importers that were already doing it.

## The fix

Both importers now use the `import_uuid` column #558 adds, and set neither `feed_name` nor `feeditem_uuid`. One dedupe concept instead of two.

With all three importers passing an `import_uuid` lookup, `run_import()`'s `feeditem_uuid` fallback had no callers left, so `$find_existing` is now required and the fallback is gone. The `$replace && $find_existing !== null` guard added in #558 became dead code and went with it.

## Migration

Installs that already imported carry the old shape, so `Bootstrap\backfill_imported_post_identity()` moves them over at boot — alongside the existing automatic `UPDATE post SET version = 1` precedent in `bootstrap_db()`. It is idempotent, so running every boot is fine.

**The guard clauses are load-bearing.** A user could have a genuinely configured feed named literally `wordpress` or `known`, and mangling their real feed posts would be far worse than the cosmetic bug this fixes. `bootstrap_db()` runs *before* `Config\load()` (`src/index.php:13,16`), so the backfill cannot consult configured feed names — SQL is the only defence available:

- `source_url IS NULL` — feed ingestion sets `source_url` from the item permalink (`src/post.php:52`); the importers never do.
- `NOT EXISTS (... other.import_uuid = post.feeditem_uuid)` — never clobber a uuid another row already owns.
- No-op when `feed_name` / `feeditem_uuid` do not exist yet (RedBean creates columns lazily, so a site that never ingested a feed has neither).

Not airtight: a feed named exactly `wordpress` or `known` whose items also stored no permalink would still be caught. No better signal exists before config load. Worth a line in the release notes.

## Behaviour change for existing sites

The "Via wordpress" / "Via known" lines disappear from already-imported posts on the next boot, and those posts start participating in webmentions and WebSub. That is the fix, but it is a visible change to live content — it belongs under **Upgrade notes**.

## Testing

Red-green throughout; `tests/Unit/KnownImportTest.php:423` asserted `feed_name === 'known'`, which pinned the bug and had to change.

New `tests/Unit/ImportIdentityBackfillTest.php`: migrates a legacy row, **leaves a genuine feed-sourced row with `feed_name = 'wordpress'` untouched because it has a `source_url`**, is idempotent across two runs, skips a uuid another row already holds, and no-ops without a `post` table. Both importers now assert empty `feed_name`/`feeditem_uuid`, a populated `import_uuid`, and — asserting the user-visible bug directly — that `Theme\link_source()` returns `''`.

1578 unit tests pass (1574 on #558). `composer lint` clean, `composer analyse` (PHPStan level 8) clean, coverage 82.72%.

## Not included

`--replace` for the WordPress and Known importers. `run_import()` now supports it for ~2 lines each and it would help re-running after fixing a conversion bug, but it is a separate concern from this fix.

https://claude.ai/code/session_01KMVaMWvgozaAViNV7iXiWG

## Added after review: `--replace` for WordPress and Known

`run_import()` already supported it, so this is wiring plus the update-in-place path. Semantics match the Lamb importer: **`--replace` re-applies everything the source describes** — body, title, slug, created/updated, draft state — onto the existing row, and local edits made since the original import are lost. Without it, an already-imported item is still counted `existed` and skipped. The use case is re-running after fixing a conversion bug, which is realistic for the Known importer's HTML→Markdown path.

Images are not re-fetched: `Import\default_image_downloader()` seeds filenames with `sha1($url)` and returns an existing file before fetching.

The two argument parsers are now one — `parse_restore_args()` delegates to `parse_import_args()` and adds only `--site-url=`, so `--dry-run` and `--replace` are spelled in a single place.

### One data-safety fix that came out of that

Unifying the parsers made flag-shaped arguments skip the path slot, which meant an unrecognised flag was **silently ignored**. That turns a typo into a real problem: `import-wordpress.php --dryrun export.xml` is one keystroke from `--dry-run`, and ignoring the bad flag ran the real, unrehearsed import the operator was explicitly trying to avoid. Before the parsers were unified this failed safe — the mistyped flag became the path and the file was not found.

An unknown flag now returns a null path, which every importer already handles by printing usage and exiting non-zero. Callers pass the flag prefixes they accept, so the shared parser can tell a caller's flag from a typo. Test asserts both `--dryrun` and a bare `-n` are refused.

1583 unit tests pass (1578 before this addition); lint, PHPStan level 8 and coverage all still clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -159,7 +159,7 @@ Each file declares a namespace; functions are called with the namespace prefix:
 RedBeanPHP (fluid mode) on SQLite. Beans are dispensed/loaded with `R::dispense`, `R::load`, `R::findOne`, `R::find`, `R::findAll`. Schema evolves automatically.
 
 **Tables used:**
-- `post` — blog posts; columns include `body`, `slug`, `title`, `description`, `transformed`, `created`, `updated`, `version`, `feed_name`, `feeditem_uuid`, `source_url`
+- `post` — blog posts; columns include `body`, `slug`, `title`, `description`, `transformed`, `created`, `updated`, `version`, `feed_name`, `feeditem_uuid`, `import_uuid`, `source_url`
 - `option` — key/value store (e.g. `site_config_ini`, `last_processed_date`, `login_fail_*` throttle counters)
 - `redirect` — automatic 301 redirects created when a post slug changes; columns: `from_slug`, `to_url`
 - `webmention` — received (inbound) webmentions; columns: `source`, `target`, `post_id`, `type`, `author`, `content`, `status`, `created`, `verified_at`
@@ -292,6 +292,7 @@ index.php
 | `$bean->updated` | Datetime string |
 | `$bean->feed_name` | Source feed name (only present for ingested feed items) |
 | `$bean->feeditem_uuid` | MD5 dedup key (only for feed items) |
+| `$bean->import_uuid` | MD5 dedup key for posts brought in by an importer (WordPress, Known, lamb export) |
 | `$bean->source_url` | Permalink of the original feed item (only for feed items; used by `link_source()`) |
 | `$bean->is_menu_item` | Truthy if the post is pinned as a menu item |
 
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -23,6 +23,21 @@ WXR was rejected: it is WordPress's schema, has no clean home for Lamb's draft/d
 
 ---
 
+## 2026-07-29 — Imported posts are your own content, not feed items
+
+**Status:** Accepted
+**Context:** The WordPress and Known importers stamped every migrated post with `feed_name = 'wordpress'`/`'known'` and a `feeditem_uuid`, borrowing the feed-ingest columns to get re-run dedup for free. Neither importer sets `source_url`, so `Theme\link_source()` took its plain-text branch and every migrated post publicly read "Via wordpress" / "Via known". Worse, `Webmention\enqueue_for_post()` and `WebSub\ping_for_post()` both bail on a non-empty `feed_name`, and the WebSub ping query excludes those rows — so a migrated post could never send a webmention or a WebSub ping, not even when edited years later. `lock_if_feed_sourced()` would also feed-lock them on first edit: exactly the trap the lamb-export importer had already refused to walk into.
+
+**Decision:** Both importers now write their dedup key to `import_uuid` — the column the lamb-export importer added — and set no feed identity at all. The `'wordpress-'`/`'known-'` prefixes stay, so the three importers keep separate dedup namespaces. Suppressing outbound notifications during an import run is still correct, and still happens the same way: `import_item()` stops at `finalize_and_store_post()`, which never calls `notify_post_subscribers()`. What is no longer true is that the suppression is permanent.
+
+With all three importers supplying their own lookup, `run_import()`'s `$find_existing` parameter became required and its `feeditem_uuid` fallback was deleted — one dedup concept instead of two.
+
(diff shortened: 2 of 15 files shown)


svandragt:
## How to test this manually

Most of this PR is invisible until you look at a post that was already imported, so the useful test needs **a database that already has WordPress or Known imports in it**. Use a copy, since the backfill rewrites rows:

```bash
cp -r data /tmp/lamb-test-data
```

### 1. Before

On the old code, open an imported post on the public site. It should show a **"Via wordpress"** or **"Via known"** line — plain text, not a link, because neither importer ever set `source_url`.

### 2. After

Boot on this branch. The backfill runs automatically from `bootstrap_db()`, so just loading any page is enough. Then check the same post:

- the "Via …" line is gone
- `feed_name` and `feeditem_uuid` are now empty, and `import_uuid` carries what `feeditem_uuid` used to
- editing the post no longer sets `feed_locked`

### 3. The case worth poking hardest

**If you have a feed genuinely configured as `wordpress` or `known`, test that first.** Its posts must be left completely alone. The backfill guards on `source_url IS NULL`, and feed ingestion sets `source_url` from the item permalink while the importers never do — that clause is the only thing standing between a real feed and having its posts stripped of their identity. `bootstrap_db()` runs before `Config\load()`, so it cannot check your configured feed names; the SQL guard is the whole defence.

The known gap: a feed named exactly `wordpress` or `known` whose items also stored no permalink would still be caught. If you can construct that, I would like to know.

### 4. Boot twice

The backfill runs on every boot by design. Second boot should change nothing.

### 5. Webmentions and WebSub

Edit and save an imported post. It should now send webmentions and ping WebSub hubs like any other post — previously `feed_name` suppressed both, permanently.

### 6. `--replace` on the two importers

```bash
devbox run -- php import-wordpress.php export.xml --replace
```

Edit an imported post first, then re-run with `--replace`: the local edit should be overwritten from the source, while the row id and `import_uuid` stay the same. Without the flag, the same re-run should report `existed=N` and change nothing.

Images are not re-fetched on a replace — `default_image_downloader()` seeds filenames with `sha1($url)` and returns an existing file before downloading — so a replace run should not re-pull your whole media library. Worth confirming on a real export, since that is a slow and visible failure if wrong.

### 7. Mistyped flags

```bash
devbox run -- php import-wordpress.php --dryrun export.xml
```

Must print usage and exit non-zero, **not** run a real import. This regressed mid-PR when the two argument parsers were unified and is now covered by a test, but it is the one failure here that would cost you a database.

## Note for the release

This changes live sites on upgrade: "Via wordpress" lines disappear from published posts and those posts start sending webmentions and WebSub pings. It belongs under **Upgrade notes**, along with the `source_url IS NULL` caveat above.

PR #587, opened 2026-08-06
Carry in-reply-to through feeds and the Micropub property paths

svandragt:
A reply's target only ever reached the theme and two feed metadata fields,
and the Micropub side of it was read-only:

- q=source omitted `in-reply-to` entirely, so a client doing
  read-modify-write on a reply saw a normal post.
- update replace/add/delete for `in-reply-to` fell through the property
  checks and returned 200, so a client believed an edit it never got had
  been saved. `set_reply_to()` now edits the front-matter key surgically,
  carrying every other key (including a hand-written `slug:` or `draft:`)
  over untouched. Adding a second target is refused with invalid_request
  rather than overwritten or dropped, since storage holds one.
- an `in-reply-to` h-cite object — which Micropub allows, and which a
  client sends when it has the parent's name to hand — reached
  matter_string() as an array, came back null, and the reply target was
  dropped. It was also missing from buildExtraProperties()'s known list,
  so the same object was dumped into the post body as a visible JSON code
  block.
- both feeds carried the relationship only in `thr:in-reply-to` and
  `_microblog.in_reply_to_url`, extensions ordinary readers ignore and
  which are not what services threading replies look for: the
  u-in-reply-to markup now travels inside the item HTML too.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011gB9uvJWJnc9EMkxFp13aF

### DIFF
--- a/docs/replies.md
+++ b/docs/replies.md
@@ -18,15 +18,28 @@ in-reply-to: https://example.com/their-post
 Great point — here's my reply.
 ```
 
-`in_reply_to` (underscore) is accepted as well. [Micropub]({{ site.baseurl }}{% link micropub.md %}) clients can send the standard `in-reply-to` property, which Lamb stores the same way.
+`in_reply_to` (underscore) is accepted as well. [Micropub]({{ site.baseurl }}{% link micropub.md %}) clients can send the standard `in-reply-to` property — either a plain URL or an embedded `h-cite` object — which Lamb stores the same way. Only one reply target is kept: a list collapses to its first entry.
 
 Remove the value from the front matter and re-save to turn the post back into a normal post.
 
 ## What it does
 
-- **On the post page** a small "In reply to …" line is shown above the content, linked to the parent and marked up with `u-in-reply-to` so Webmention receivers treat it as a reply.
+- **On the post page and in listings** a small "In reply to …" line is shown above the content, linked to the parent and marked up with `u-in-reply-to` so Webmention receivers treat it as a reply.
 - **Atom feed**: emits `<thr:in-reply-to ref="…" href="…" />` (the `http://purl.org/syndication/thread/1.0` thread extension).
 - **JSON feed**: emits `_microblog.in_reply_to_url` (the micro.blog reply convention).
+- **Both feeds** also carry the same "In reply to …" line inside the item's HTML content, because the two metadata fields above are extensions most readers ignore, and services that thread replies look for the `u-in-reply-to` markup itself.
+
+## Editing a reply over Micropub
+
+A Micropub client sees the reply target in a `q=source` response as the `in-reply-to`
+property, and can change it with an update:
+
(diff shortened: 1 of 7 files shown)


svandragt (commented):
Reviewed the diff and ran the suite on this branch — 1716 tests, green. The four bugs in the description are real and the fixes read correctly; the h-cite handling and the surgical front-matter edit are both the right calls. Two findings, one of which is the same class of bug this PR sets out to fix.

## 1. `add` returns 200 for a value it cannot use — `src/micropub.php:541`

```php
$target = $this->replyTargetUrl($values[0] ?? null);
if ($target === null) {
    return true;   // <- 200, nothing saved
}
```

`applyReplace()` returns `false` for exactly this input (an h-cite with no `url`), and the PR description calls out why: a 200 for an edit the storage never made reads to the client as "saved". `applyAdd()` takes the opposite path.

Confirmed on this branch:

```
add {in-reply-to: [{type: [h-cite], properties: {name: ["No url here"]}}]}
  -> updateCallback returned true, stored in_reply_to = ''
```

The same payload through `replace` returns `invalid_request`.

The empty list has to stay a no-op, so it isn't a bare `return false`:

```php
if ($values === []) {
    return true;
}
$target = $this->replyTargetUrl($values[0] ?? null);
if ($target === null) {
    return false;
}
```

## 2. `set_reply_to()` drops sibling keys under a nested parent — `src/post.php:479`

The key regex allows leading whitespace, so it matches an *indented* `in-reply-to`, and the continuation-line rule then eats the rest of that nested block.

```
---                          ---
meta:                        meta:
  in-reply-to: https://old/x     ->    title: T
  other: keep-me               in-reply-to: 'https://new.example/post'
title: T                     ---
---
```

`other: keep-me` is gone and `meta:` is left with a null value. The docblock promises every other key is "carried over verbatim, because the Micropub update path that calls this must not silently drop a hand-written `slug:` or `draft:`" — this is that failure mode, one level down.

Anchoring the key to column zero fixes it, since the front matter Lamb writes is flat:

```php
preg_match('/^in[-_]reply[-_]to[ \t]*:/i', $line)
```

Low severity — nothing in Lamb writes nested front matter, so it only bites a hand-written file. Worth the one-character fix given the docblock's promise.

## Nits, no action needed

- `beanToMf2Properties()` returns the target as a plain URL string, so an h-cite that came in as `{url, name}` comes back out of `q=source` without the name. Correct given storage holds one URL — just means a strict read-modify-write client loses the citation metadata it sent.
- The `addChild()` comment in `feed.php` is accurate: `the_reply_context()` runs the URL through `escape()`, and an already-encoded `&amp;` survives the round-trip intact. Worth knowing that a *bare* `&` would not — `addChild()` silently deletes it rather than escaping — so that comment is load-bearing for anyone refactoring the helper.

Issue #595, opened 2026-08-09
Follow up on micropub-adapter PHP 8.4 deprecation (upstream issue #4)


svandragt:
Track https://github.com/Taproot/micropub-adapter/issues/4 and decide what we do if it goes unanswered.

## Background

`taproot/micropub-adapter` v0.1.3 declares an implicitly nullable parameter, which PHP 8.4 deprecated:

```php
// src/MicropubAdapter.php:224
public function sourceQueryCallback(string $url, array $properties = null)
```

```
Deprecated: Taproot\Micropub\MicropubAdapter::sourceQueryCallback():
Implicitly marking parameter $properties as nullable is deprecated,
the explicit nullable type must be used instead
```

The fix upstream is one character — `?array`. Valid since PHP 7.1, and their composer.json already requires `>=7.3`, so it costs them no constraint bump. It's the only occurrence in `src/`.

## Why we care

The notice prints at class-load time, before any application output. Where `display_errors` is on, that means "headers already sent" and hard failures rather than a warning — running the suite under devbox PHP 8.4 locally produced 19 errors and 8 failures for this reason. CI is unaffected because `display_errors` is off there, so it passes on 8.4 and 8.5 while a developer machine falls over. That split is the annoying part: the failure looks like a real regression until you trace it.

Not urgent — production has `display_errors` off.

## Decide

Upstream has been quiet since July 2023, so we should not assume a response. If #4 gets no traction:

- [ ] Open a PR upstream ourselves — one-line change, cheapest path if anyone is still merging
- [ ] Composer patch via `cweagans/composer-patches` — keeps us on the released package, adds a dependency and a patch file to maintain
- [ ] Fork and point the repository at ours — full control, we own the maintenance
- [ ] Vendor the adapter into `src/` — no dependency at all, but we inherit the whole file
- [ ] Accept it and set `display_errors=off` in the local dev PHP config — hides the symptom, keeps the trap for the next person

## Check back

Revisit when #4 has been open a month with no maintainer reply (around 2026-09-09), or sooner if it blocks work.

Related: #573 raised the PHP floor to 8.4, which is what surfaced this.

svandragt:
@svandragt Triage: this is a decision issue with a check-back date of ~2026-09-09 (upstream #4 idle since 2023). Which path do you want when it stays quiet — upstream PR, composer-patch, fork, or vendor? Until you pick one it stays parked rather than ready. No action needed before the check-back date; flagging so it is a conscious wait, not a stall.

PR #611, opened 2026-08-16
Support installs served from a subdirectory (#580)

### LINKED ISSUE #580: Micropub rejects all tokens on a subdirectory install (site_url path is stripped)
## Summary

`Lamb\Config\canonical_site_url()` normalises the configured site URL to `scheme://host[:port]`, discarding any path. On an install served under a subpath (e.g. `https://example.com/blog`), the author's IndieAuth identity *is* `https://example.com/blog`, so the `me` comparison in `verify_token()` can never match and every Micropub token is refused with `me_mismatch`.

## Where

- `src/config.php:368` — `canonical_site_url()` returns `$scheme . '://' . host . $port`; the parsed path is never reappended.
- `src/micropub.php:161` — `rtrim($data['me'], '/') !== rtrim($expected, '/')` → mismatch, token rejected.
- `tests/Unit/CanonicalSiteUrlTest.php:33` asserts the current behaviour (`http://example.com:8747/blog` → `http://example.com:8747`), so this is deliberate today rather than an accident.

## Reproduction

With `site_url = https://example.com/blog` and a token whose `me` is `https://example.com/blog`:

```
expected: https://example.com
me:       https://example.com/blog
→ mp_log('token_verify', ['reason' => 'me_mismatch', ...]) and 403
```

Confirmed by calling `canonical_site_url()` directly: `['site_url' => 'https://EXAMPLE.com/blog/']` returns `https://example.com`.

There is no configuration that works around it — the env override `LAMB_SITE_URL` goes through the same normalisation, so a subpath identity cannot be expressed at all.

## Why it is not just a normalisation bug

The path is stripped for a reason worth preserving: `canonical_site_url()` is the one value trusted for a security decision, and a loose parse there widens what an author can accidentally configure into the identity comparison (query strings, fragments, traversal segments, mixed trailing-slash forms). Any fix needs to keep the value strictly normalised and comparison-stable, not merely stop discarding the path.

Also note `ROOT_URL` (`src/index.php:24`) has the same shape today, so preserving a path in the canonical value would start to diverge from the derived request fallback — every consumer that concatenates `ROOT_URL . '/...'` (feeds, discovery, webmention, export) needs checking as part of any change.

## Options

1. **Support subpaths properly.** Preserve a normalised path (strip query/fragment, collapse duplicate slashes, reject `..`, drop the trailing slash) and make `ROOT_URL` and every absolute-URL builder path-aware. Correct, but touches most consumers of `ROOT_URL`.
2. **Document the limitation and fail loudly.** Keep root-only identities, but reject a `site_url` carrying a path at save time in `/settings` with a clear message, and say so in `docs/micropub.md` + `docs/site-configuration.md`. Cheap, and turns a silent 403 into an explanation.
3. **Status quo.** Leave it; subpath installs silently cannot use Micropub.

Option 2 is the smallest change that stops the silent failure; option 1 is the real fix if subpath installs are in scope for the product at all — that product call is the thing to decide first.

## Acceptance criteria

- A subpath `site_url` either works end-to-end for Micropub, or is rejected at configuration time with a message naming the limitation.
- No path-bearing value can reach the `me` comparison without being normalised to a single canonical form.
- Behaviour is covered by unit tests alongside the existing `CanonicalSiteUrlTest` cases, and the chosen behaviour is documented in `docs/micropub.md`.

svandragt:
**Draft, and may not be merged.** Opened so the work is reviewable and the trade-off is visible, not because it should land. See "Is this worth it?" below — there is a much smaller alternative.

Fixes #580.

## The bug

An install under `https://example.com/blog` cannot use Micropub at all. `canonical_site_url()` normalised the configured site URL down to `scheme://host[:port]` and threw the path away, so the author's IndieAuth identity could never match the `me` in a token. Every request was refused with `me_mismatch`, and no configuration worked around it — `LAMB_SITE_URL` went through the same normalisation, so a subpath identity could not be expressed at all.

## Why the fix is this big

Keeping the path makes the base something the whole app has to apply, in both directions.

**Taking it off.** `strip_base_path()` removes the base from an incoming request before anything segments it, or the router reads `blog` as the action. `find_post_by_path()` needs the same treatment, because its callers hand it the path out of a permalink — without that a Micropub client cannot update or delete the post it was just given, and every inbound webmention is rejected as an unknown target. That second one was found in review, and it is the same bug as #580 wearing a different hat.

**Putting it on.** `app_path()` is the counterpart. Links, form actions, redirect targets, configured menu items and `[redirections]`, and the two client-side `fetch()` calls all go through it. Absolute URLs needed no change — they concatenate onto `ROOT_URL`, which now carries the base.

**Content.** Post bodies keep storing `/assets/…` unchanged so content stays portable across a domain change; the base goes on at render time via a new `Theme\the_content()`. That is now the single seam a post body passes through, and it absorbed the `anchor_headings()` call each theme was making separately.

**Normalisation.** A base path feeds a security comparison, so it is normalised hard: one canonical spelling, a narrow charset, and a refusal rather than a decode for anything else — `..` and `%2e%2e` cannot reach different answers if neither is allowed through.

## Is this worth it?

Honest summary of the cost, since the answer may be no:

- 36 files. The base path is a new concept every future link, redirect and fetch has to respect, and nothing enforces that — the review found four places I had already missed.
- Nobody is known to run Lamb under a subdirectory.
- Issue #580 lists a second option: keep root-only identities, but reject a path-bearing `site_url` at save time with a message naming the limitation, and document it. Roughly 20 lines instead of this. It turns a silent 403 into an explanation without inventing a base-path concept.

The main thing in this branch's favour: at a domain root the base is `''`, and `app_path()`, `strip_base_path()` and `the_content()` are all identity in that case. So the risk to existing installs is low even though the diff is wide.

## Checks

- 1692 unit tests, 28 new. Every one written failing first.
- JS 40/40, PHPStan clean, phpcs clean.
- The suite's 21 errors / 7 failures are the pre-existing #595 deprecation baseline, byte-identical on `main`.

**Not verified on a real subdirectory install.** There is no acceptance coverage, because that needs a web-server rewrite the test suite does not set up. This is unit-level only.

## Known limitation

A hand-pasted internal path that already includes the base gets it twice: `/blog/assets/x.webp` renders as `/blog/blog/assets/x.webp`. It cannot be detected away — a post slugged `blog` on a `/blog` install is stored as `/blog` and must become `/blog/blog`. Pinned in a test with the reasoning so it is not "fixed" into a regression.

https://claude.ai/code/session_01CYmuANUvADwwYmg2AbgZoQ

### DIFF
--- a/docs/micropub.md
+++ b/docs/micropub.md
@@ -29,6 +29,17 @@ Do not derive it from the request — that is exactly what this setting exists t
 avoid. You can also supply it as the `LAMB_SITE_URL` environment variable, which
 takes precedence over the setting.
 
+If your site is served from a subdirectory, include the path:
+
+```ini
+site_url = https://example.com/blog
+```
+
+Your IndieAuth identity is the whole address, path and all, so a host-only value
+here never matches the identity in your token and every request is refused. See
+[Installing under a subdirectory](site-configuration.md#installing-under-a-subdirectory)
+for what a base path may contain.
+
 ### 2. Add `rel="me"` identity links
 
 IndieAuth verifies who you are by checking that your site links to your profiles and those profiles link back. Add a `[me]` section to your site configuration at `/settings`:
--- a/docs/site-configuration.md
+++ b/docs/site-configuration.md
@@ -157,6 +157,29 @@ You can also set it outside the settings page with the `LAMB_SITE_URL` environme
 variable, which takes precedence over the INI value — handy when the same
 configuration is deployed to more than one hostname.
 
+### Installing under a subdirectory
+
+To serve Lamb from a subdirectory, include the path in `site_url`:
+
+```ini
+site_url = https://example.com/blog
+```
+
+Lamb reads the path as the install's base. Links, forms, redirects and uploads
+all go through it, and the base comes back off an incoming request before the
+router looks at it. Point your web server's rewrite at the same prefix.
+
+Set `site_url` before you use a subdirectory install. It is the only place Lamb
+learns its own base — without it, an install under a path behaves as though it
+sits at the domain root.
+
+Keep the value simple, because Lamb compares it against the identity in your
+Micropub tokens and both spellings have to match. Use letters, digits, dots,
+dashes, underscores and tildes in the path. Lamb drops a trailing slash, collapses
+repeated slashes, and ignores any query or fragment. It refuses a path holding
+anything else — a space, a percent-encoded character, or a `.` or `..` segment —
+and falls back to behaving as though no site URL is set.
+
 ## Related
 
 * [Setting up Cross-Posting]({{ site.baseurl }}{% link cross-posting.md %}#setup) requires site configuration changes.
--- a/src/config.php
+++ b/src/config.php
@@ -29,6 +29,8 @@ theme = 2026
 site_title = My Microblog
 
 ;; The canonical URL of your site, e.g. https://example.com (no trailing slash).
+;; Serving from a subdirectory? Include the path: https://example.com/blog — it is
+;; how Lamb learns its own base, and part of the identity your tokens are issued for.
 ;; Set this if you use Micropub: it is how Lamb tells that an IndieAuth token was
 ;; issued for your identity rather than someone else's, so the endpoint refuses
 ;; every token while it is unset. It also pins the absolute URLs in feeds and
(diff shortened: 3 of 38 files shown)


svandragt:
not planned

PR #612, opened 2026-08-16
Warn when site_url carries a path instead of failing silently (#580)

svandragt:
**Draft.** This is the small alternative to #611 — the two are mutually exclusive answers to the same issue. Pick one and close the other.

Refs #580.

## The problem, unchanged

An install at `https://example.com/blog` cannot use Micropub. `canonical_site_url()` narrows the configured value to `scheme://host[:port]`, so the path is dropped. That value is the author's IndieAuth identity, so Lamb compares `https://example.com` against a token issued for `https://example.com/blog`, and refuses every one with `me_mismatch`.

The worst part is not the refusal. It is that the settings page says "Settings saved successfully", and the failure appears somewhere else entirely, in a client that will not connect.

## What this does

Says so at the point of saving, and documents the limitation. It does not add subdirectory support — an install under a path still cannot work.

`shape_warnings()` already exists for exactly this shape of problem: a setting that saved and will be ignored, reported back to the author on the settings page. The new warning names the path, says it is dropped, says Micropub will refuse every token as a result, and points at the fix — serve the site from its own domain or subdomain.

A `site_url` that `canonical_site_url()` cannot read at all is left alone. That case already has its own handling and its own message, and a second differently-worded complaint is noise.

## One judgement call

This widens `shape_warnings()` past pure shape, which its own docblock had ruled out: "Whether a value is a usable timezone, URL or theme name is each reader's business and each already has its own fallback."

`site_url` is the exception, and the docblock now says why. Every other narrowed value degrades where the author can see it. This one is silently narrowed and the damage surfaces in a different subsystem, with nothing connecting the two.

## Compared with #611

| | This PR | #611 |
|---|---|---|
| Files | 4 | 36 |
| Lines | +95, −0 | +593, −76 |
| Subdirectory installs | Still unsupported, but explained | Supported |
| New concepts | None | A base path every future link, redirect and fetch must respect |
| Ongoing cost | None | The review found four places already missed |

#611 is the real fix if subdirectory installs are in scope for Lamb. This one is the right answer if they are not, and it is worth noting that nobody is known to run one.

Whichever way it goes, one thing from #611 is worth keeping either way: `find_post_by_path()` has the same base-path bug, so on a subpath install Micropub update and delete, and every inbound webmention, fail to resolve their target. That is moot if subdirectories stay unsupported, but it is the same root cause wearing a different hat.

## Checks

- 3 new unit tests, written failing first.
- PHPStan clean, phpcs clean.
- The suite's 21 errors / 7 failures are the pre-existing #595 deprecation baseline, byte-identical on `main`.

https://claude.ai/code/session_01CYmuANUvADwwYmg2AbgZoQ

### DIFF
--- a/docs/micropub.md
+++ b/docs/micropub.md
@@ -29,6 +29,10 @@ Do not derive it from the request — that is exactly what this setting exists t
 avoid. You can also supply it as the `LAMB_SITE_URL` environment variable, which
 takes precedence over the setting.
 
+Give the domain on its own, with no path. Lamb serves from the root of a domain,
+so a value like `https://example.com/blog` cannot work — see
+[Troubleshooting](#troubleshooting).
+
 ### 2. Add `rel="me"` identity links
 
 IndieAuth verifies who you are by checking that your site links to your profiles and those profiles link back. Add a `[me]` section to your site configuration at `/settings`:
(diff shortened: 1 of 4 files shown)


svandragt:
This is a better representation of how it is currently, not presenting the blogger with a footgun.

PR #615, opened 2026-08-16
Unblock the release image scan

svandragt:
The `Release artifacts` Trivy gate has failed on every release since 0.12.0 — 3 August and again on 0.13.0-rc1 — so no image has reached `ghcr.io/svandragt/lamb` since 0.12.0. Two independent causes, both fixed here.

## linux-libc-dev

`dunglas/frankenphp:php8.4` ships a C/C++ toolchain so `install-php-extensions` can compile, and that pulls in `linux-libc-dev`. Debian attaches every kernel CVE to those headers, so the scan reported 18 HIGHs (`6.12.100-1`, fixed in `6.12.101-1`) for a kernel the container never runs — it uses the host's.

Purging the package once the extensions are built removes the whole recurring class, and takes `g++`, `libstdc++-14-dev` and `libc6-dev` with it. Nothing at runtime wants them. Upgrading the package instead would have cleared today's 18 and left the next batch to block the next release.

## Two Go advisories in the FrankenPHP binary

`GHSA-r277-6w6q-xmqw` (kin-openapi fail-open auth bypass) and `GHSA-hrxh-6v49-42gf` (gRPC-Go xDS RBAC) are modules linked into the shipped `frankenphp` binary. Lamb cannot patch them, and no newer base image tag has them fixed yet. Neither is reachable through `.docker/Caddyfile.release`, which serves static files and PHP and turns on no OpenAPI validation, gRPC or xDS.

They go in `.trivyignore` with the versions that would clear them, so the next release can drop the lines and check.

## Verified locally

Built the image and scanned it with the same Trivy settings the workflow uses:

```
Target                              Type             Vulnerabilities
img.tar (debian 13.6)               debian           0
app/vendor/composer/installed.json  composer-vendor  0
usr/local/bin/frankenphp            gobinary         0
```

The built image serves `/` and `/login` with a `200`, runs as `www-data` (uid 33), still has `curl` for the healthcheck, and loads `gd`, `gettext` and `pdo_mysql`.

https://claude.ai/code/session_01CeZkmJnKHa8tgLV6X9xJcj

### DIFF
--- a/.docker/Dockerfile.release
+++ b/.docker/Dockerfile.release
@@ -16,6 +16,13 @@ LABEL org.opencontainers.image.source="https://github.com/svandragt/lamb"
 # gd powers WebP upload conversion
 RUN install-php-extensions gettext pdo_mysql gd
 
+# The base image ships a C/C++ toolchain so install-php-extensions can compile;
+# nothing needs it once the extensions are built. Dropping it also drops
+# linux-libc-dev, whose kernel CVEs blocked every release scan — the container
+# runs the host's kernel, so those headers were never the thing at risk.
+RUN apt-get purge -y --auto-remove linux-libc-dev \
+    && rm -rf /var/lib/apt/lists/*
+
 RUN cp "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
 
 # Images are converted to WebP and resized server-side, so large photo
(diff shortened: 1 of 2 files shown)


svandragt:
Landed on `main` as d5f6345 — the squash-merge created the commit but the API call 502'd before it recorded the merge on this PR, so it stayed open. `.docker/Dockerfile.release` and `.trivyignore` on `main` are identical to this branch, and both are on `release` in 0.13.0-rc1.

PR #649, opened 2026-08-21
Require the database driver Lamb actually connects through

svandragt:
`composer.json` required `ext-pdo_mysql` and `ext-sqlite3`, and did not require `ext-pdo_sqlite`. The platform check therefore fails on exactly the wrong hosts: it refuses one Lamb runs on, and passes one it cannot run on.

## Why those requirements are wrong

`bootstrap_db()` calls `R::setup("sqlite:…")`, and RedBeanPHP picks its query writer from the DSN scheme (`Facade.php`): a `sqlite` DSN instantiates `QueryWriter\SQLiteT` and never loads `QueryWriter\MySQL`. The connection itself is a `PDO` against the **sqlite** driver. RedBeanPHP v5.7.6 requires nothing but `php >=5.3.4` — the "required by the database library" note in the docs has no basis in the library.

Nothing in `src/` touches the sqlite3 *extension* either. The one apparent hit, `sqlite_master` in `ensure_post_columns()`, is a SQL table name.

## Both failure directions, reproduced

Each against a PHP with the extension in question actually removed.

**A host with `pdo_sqlite` but no MySQL extensions** — Lamb works there:

```
$ PHP_INI_SCAN_DIR=…/no-mysql php extcheck.php
stored id=1 read=hello
MySQL writer class loaded: false
SQLiteT writer loaded: true
```

…and `composer install` refuses it:

```
ext-pdo_mysql n/a  svandragt/lamb requires ext-pdo_mysql (*)  missing
ext-sqlite3   n/a  svandragt/lamb requires ext-sqlite3 (*)    missing
```

Debian's `php8.4-mysql` was being installed for nothing.

**A host with no `pdo_sqlite`** — every declared requirement reports `success`, and the first write dies:

```
PHP Fatal error:  Uncaught PDOException: Could not connect to database (?).
  in RedBeanPHP/Driver/RPDO.php:473
```

RedBeanPHP swallows PDO's own "could not find driver", so the operator gets no hint that a driver is missing.

After this change the first host installs clean (all `success`) and the second is refused with `requires ext-pdo_sqlite`.

## `ext-mbstring`

Added for the same reason: `src/` calls `mb_strimwidth()`, `mb_substr()` and `mb_check_encoding()` directly, and only happened to be covered because `erusev/parsedown` and `phiki/phiki` declare it. Every host that can install Lamb today already has it, so this refuses nothing new — it just attributes the requirement to the code that has it.

`ext-dom` stays undeclared: `league/html-to-markdown` requires it, and `check-platform-reqs` already reports it.

## Docs

`docs/index.md` and `docs/local-php-setup.md` both repeated the pdo_mysql claim; both now name the driver the code opens, and the `apt install` line drops `php8.4-mysql` (`php8.4-sqlite3` already ships `pdo_sqlite`).

## Test

`tests/Unit/PlatformRequirementsTest.php` pins the pairing rather than restating the list: every declared `ext-*` has to name something the source calls into, the unused database extensions must stay out, and a store must leave `QueryWriter\MySQL` unloaded. The `pdo_sqlite` probe checks `PDO::getAvailableDrivers()` rather than `class_exists('PDO')`, which would also pass with no driver at all.

Ran against the pre-fix manifest, 4 of the 6 assertions fail; after, all 6 pass. On a host without `pdo_sqlite` the last one fails as it should. `composer validate` is clean (the lock's platform block is regenerated).

---


svandragt:
Pushed a second commit taking `pdo_mysql` out of both images, so the claim isn't left half-applied. Same reasoning: `install-php-extensions` was installing the MySQL driver for a data layer that never loads `QueryWriter\MySQL`.

Neither image has ever installed `pdo_sqlite`, and both have always served a SQLite database — so `dunglas/frankenphp:php8.4` already ships the driver Lamb connects with, and the line only needs the unused extension dropped, not replaced.

Touching `.docker/**` turns on the `image-scan` job, so the release image build is exercised by this PR's CI. The dev image (`.docker/docker-compose.yml` only) has no CI coverage either way.

One related thing I did **not** change: the dev image also installs `ext-yaml`, which nothing uses — `symfony/yaml` is pure PHP and `src/` only calls `Yaml::dump()`/`Yaml::parse()`. It belongs in its own change rather than riding along with a database-driver fix.

---

svandragt:
## Tested: the `pdo_mysql` removal is a regression — I've pushed a fix to this branch

I reproduced the original requirement against a PHP built with `pdo_sqlite` and **without** `pdo_mysql` (a custom ini dropping only that extension). The premise that "no MySQL extension is involved" doesn't hold — the requirement was never about the query writer.

### What actually happens without `pdo_mysql`

RedBeanPHP v5.7.6's `Functions.php` (loaded via Composer's `files` autoload, so it runs on **every** bootstrap) dereferences the MySQL PDO constant at load time:

```php
// vendor/gabordemooij/redbean/RedBeanPHP/Functions.php:136
if (defined('Pdo\Mysql::ATTR_INIT_COMMAND')) {
    define('RB_PDO_MYSQL_ATTR_INIT_COMMAND', Pdo\Mysql::ATTR_INIT_COMMAND);
} else {
    define('RB_PDO_MYSQL_ATTR_INIT_COMMAND', \PDO::MYSQL_ATTR_INIT_COMMAND);  // ← fatal
}
```

Both forms of that constant come only from `ext-pdo_mysql`. Without it, `defined()` is false, the `else` branch runs, and autoloading fatals **before any request logic**:

```
Fatal error: Uncaught Error: Undefined constant PDO::MYSQL_ATTR_INIT_COMMAND
  in vendor/gabordemooij/redbean/RedBeanPHP/Functions.php on line 140
```

The stack shows it dying inside `require 'vendor/autoload.php'` → the redbean `files` autoload — so `R::setup("sqlite:…")` is never even reached. Merging this as-is (composer.json + both Dockerfiles) would break Lamb on every host.

### Why the "Lamb works on a host with no MySQL extensions" repro passed anyway

The constant `PDO::MYSQL_ATTR_INIT_COMMAND` is provided by **`pdo_mysql`**, not `mysqli`/`mysqlnd`. Verified:

| ini variant | `pdo_mysql` | `mysqli` | `defined('PDO::MYSQL_ATTR_INIT_COMMAND')` |
|---|---|---|---|
| drop mysqli+mysqlnd, keep pdo_mysql | n* | n | — |
| drop pdo_mysql, keep mysqli | n | y | **MISSING** |

So a repro that removed "MySQL extensions" but left `pdo_mysql` installed would never trip the fatal — which is the likely gap here.

### What was right in this PR (kept)

- **`ext-pdo_sqlite`** — correct; it's the driver the sqlite DSN opens and it was undeclared. Confirmed a store works with it present.
- **`ext-mbstring`** — correct; `src/` calls `mb_*` directly.
- **dropping `ext-sqlite3`** — correct; no `SQLite3`/`sqlite3_*` API usage in `src/` (the `sqlite_master` hit is a SQL table name).
- **`MySQL writer stays unloaded`** — true, and worth keeping as a test: baseline store gives `MySQL writer loaded: false`. It just isn't the same thing as the extension requirement.

### The fix I pushed (9f73019)

Restores `ext-pdo_mysql` in `composer.json` + both images, keeps all the genuine improvements above, rewrites the docs to state the *real* reason (the autoloaded constant, not "the database library needs MySQL"), and rewrites the test to pin both directions — `pdo_mysql` required (probing the exact constant RedBean dereferences), `pdo_sqlite` required, `mysqli`/`sqlite3` stay out, store never loads the MySQL writer.

Green here: `PlatformRequirementsTest` 6/6, `composer validate` clean, `composer check-platform-reqs` passes, PHPStan level 8 clean.

svandragt:
You're right, and your commit is the correct fix. I checked it against the vendored source rather than taking it on trust, and it holds up exactly as described.

`gabordemooij/redbean` v5.7.6 `composer.json`:

```json
"autoload": { "files": ["RedBeanPHP/Functions.php"], "psr-4": { "RedBeanPHP\\": "RedBeanPHP" } }
```

`Functions.php:137-140`, and loading it directly:

```
with pdo_mysql:  loaded ok, RB_PDO_MYSQL_ATTR_INIT_COMMAND=1002
without:         PHP Fatal error: Uncaught Error: Undefined constant
                 PDO::MYSQL_ATTR_INIT_COMMAND in RedBeanPHP/Functions.php:140
```

**What I got wrong.** My "Lamb works fine without any MySQL extension" demonstration booted RedBeanPHP through a hand-rolled loader in my verification harness that pulls in the classes directly and never includes `Functions.php` — only Composer's `files` autoload does. So I measured a boot path the real application doesn't use, proved the query-writer half, and generalised from it to a conclusion that doesn't follow. The requirement was never about the writer, as you say.

**The part that matters most:** my second commit took `pdo_mysql` out of both images, which would have broken the released container at boot. `image-scan` builds the release image but never runs it, so green CI could not have caught that. Thanks for catching it before it shipped.

I've verified your head (`9f73019`) on every axis:

| | result |
| --- | --- |
| `composer validate` | clean |
| host with `pdo_sqlite`, no MySQL | now correctly **refused** on `ext-pdo_mysql` |
| host with no `pdo_sqlite` | still refused on `ext-pdo_sqlite` |
| both Dockerfiles | `pdo_mysql` restored |
| net change vs main | add `pdo_sqlite`, add `mbstring`, drop `sqlite3` — the three that were genuinely wrong |

I also checked whether anything else could bite the same way, since the failure class is "a `files`-autoloaded package dereferences an extension symbol at load, before any of your code runs". Three production packages have a `files` autoload — `gabordemooij/redbean`, `symfony/deprecation-contracts`, `symfony/polyfill-ctype`. The other two are pure-PHP shims, and that `PDO::MYSQL_ATTR_INIT_COMMAND` line is the only load-time extension reference in `Functions.php`. So there is no second instance of this hiding in the current dependency set.

I've fixed my harness to include `Functions.php`, so it now fails where the real app would instead of booting.

One suggestion, entirely your call and not something I'd push here: `image-scan` builds the image but never starts it, which is precisely why this got through. Running the built image and hitting its `HEALTHCHECK` endpoint once would turn "the image builds" into "the image boots". Happy to open that separately if you want it.

---

PR #653, opened 2026-08-21
Do not report a stored file the media endpoint failed to store

svandragt:
`respond_micropub_media()` ignores `move_uploaded_file()`'s return value and then answers `201` with a `Location` header, so a failed write is reported to the client as a successful upload:

```php
$filename = \Lamb\Response\store_webp_copy($file['tmp_name'], $ext, $uploadDir, $seed);
if ($filename === null) {
    $filename = $seed . ".$ext";
    move_uploaded_file($file['tmp_name'], $uploadDir . '/' . $filename);   // <- result dropped
}
...
http_response_code(201);
header('Location: ' . $url);
```

`/upload` has the same store-or-fall-back-to-move structure and checks the same call, answering 500. Measured, for a `.gif` whose move fails: `/upload` returns `"Move upload error: …"` with a 500; the media endpoint returned 201.

## The fallback is the ordinary path, not a rare one

`store_webp_copy()` only converts JPEG and PNG. Across the whole upload allowlist:

| ext | `store_webp_copy()` | | ext | | | ext | |
| --- | --- | --- | --- | --- | --- | --- | --- |
| jpg | `seed.webp` | | gif | `null` | | mp4 | `null` |
| jpeg | `seed.webp` | | webp | `null` | | webm | `null` |
| png | `seed.webp` | | avif | `null` | | mov | `null` |

So every video, gif, webp and avif uploaded through the media endpoint is stored by the unchecked move. On a full disk or a read-only assets mount the client is handed a URL for a file that does not exist — and the media endpoint exists precisely so the client can put that URL into a post, where it 404s for good.

`server_error` follows OAuth 2.0 (RFC 6749 §4.1.2.1); Micropub defines no code of its own for a 500.

## Verification, and the test I did not write

There is no new test here, and that is a real gap rather than an oversight: making `move_uploaded_file()` fail requires a genuine registered upload, which cannot be arranged in-process, and the endpoint's auth path introspects the token against the configured token endpoint over HTTP.

What I measured instead: the extension table above, from `store_webp_copy()` itself; that `move_uploaded_file()` on a non-upload returns `false` and creates nothing; and the sibling endpoint's answer for the identical failure.

Once #651's batch rework lands, both call sites can share one store-the-uploaded-file helper — which would also be unit-testable — instead of repeating the pair. Kept out of this PR so the two don't conflict on the same lines.

---

### DIFF
--- a/src/micropub.php
+++ b/src/micropub.php
@@ -1579,7 +1579,14 @@ function respond_micropub_media(): void
     $filename = \Lamb\Response\store_webp_copy($file['tmp_name'], $ext, $uploadDir, $seed);
     if ($filename === null) {
         $filename = $seed . ".$ext";
-        move_uploaded_file($file['tmp_name'], $uploadDir . '/' . $filename);
+        // Checked, the way /upload checks it. Everything store_webp_copy()
+        // declines — gif, webp, avif and every video — lands here, so this is
+        // the ordinary path for a video upload, not a rare one. Answering 201
+        // with a Location for a file that was never written hands the client a
+        // URL it embeds in a published post, where it 404s forever.
+        if (!move_uploaded_file($file['tmp_name'], $uploadDir . '/' . $filename)) {
+            micropub_error(500, 'server_error', 'Could not store the uploaded file.');
+        }
     }
 
     // The media endpoint hands this URL back to an external Micropub client, so


svandragt:
needs tested locally against a variety of upload types.

svandragt:
Tested locally against every allowed upload type, driving the real `respond_micropub_media()` over HTTP with real multipart uploads.

**Rig:** two loopback PHP servers — one serving `/micropub-media` and the assets tree, one standing in as the IndieAuth token endpoint (`introspectToken()` uses the unguarded fetcher, so loopback works). Real bearer token, real introspection, real `$_FILES`, so `move_uploaded_file()` behaves as it does in production. Fixtures: GD-generated jpg/jpeg/png/gif/webp/avif, minimal ISOBMFF and EBML files for mp4/mov/webm, plus three that should be refused. The returned `Location` is fetched back to check the file is really there.

**Healthy destination, `main` — all nine store and resolve:**

| type | code | stored as | GET |
| --- | --- | --- | --- |
| photo.jpg / photo.jpeg / shot.png | 201 | `…webp` | 200, 78 B |
| anim.gif | 201 | `…gif` | 200, 50 B |
| pic.webp | 201 | `…webp` | 200, 78 B |
| pic.avif | 201 | `…avif` | 200, 313 B |
| clip.mp4 | 201 | `…mp4` | 200, 160 B |
| clip.webm | 201 | `…webm` | 200, 140 B |
| clip.mov | 201 | `…mov` | 200, 148 B |
| notes.txt / shell.php | 400 | — | `Unsupported file type.` |
| evil.png (HTML bytes) | 400 | — | `File contents do not match its type.` |

Every pass-through type is byte-identical to what was uploaded (sha256 matched for mp4, webm, mov, gif, avif), and the three converted files are real 24×16 WebP images.

**Unwritable destination** — the month directory made genuinely unwritable with `chattr +i`, so `is_dir()` still passes and `get_upload_dir()` succeeds but nothing can be created in it. This is the read-only-mount / permissions case the PR is about:

```
main    all 9 allowed types -> 201 Created, Location 404s
#653    all 9 allowed types -> 500 {"error":"server_error", …}
```

Then with the destination writable again, `#653` reproduces the healthy table above exactly — no regression on any type, and the three refusals are unchanged in all four runs.

**One correction to the PR description.** I wrote that the unchecked move was the path for "every video, gif, webp and avif". In the actual failure scenario it is *all nine*: an unwritable destination also defeats `store_webp_copy()`, so JPEG and PNG fall through to the same unchecked `move_uploaded_file()` rather than converting. The bug was broader than I described.

I have not marked this ready for review — that is yours to do. I also did not add an automated test: the rig needs two HTTP servers and a filesystem the process cannot write to, which is beyond what the unit suite sets up. If you would like it as a Playwright e2e case instead, say the word and I will write it.

---

svandragt:
Closing as redundant: main already carries this exact guard via 00c9935 ("Check move_uploaded_file() in the Micropub media endpoint"). This PR was branched from an older base (8e2fc24) and would only conflict on the same lines.

The one genuine improvement here is the error code — a failed disk write is a server fault, so it should be `server_error` (RFC 6749 §4.1.2.1), not the `invalid_request` main currently ships (a 400 client-error code, §5.2). Following that up as a one-line fix to main instead.

Issue #672, opened 2026-08-21
Consolidate media-endpoint and /upload store-or-fallback into one helper


svandragt:
`respond_micropub_media()` and the `/upload` handler repeat the same store-the-WebP-copy-then-`move_uploaded_file()` fallback structure. #653 fixed a dropped return value in one copy but deliberately left the two call sites separate to avoid conflicting with #651's batch-upload rework.

Once #651 lands, fold both into a single, unit-testable helper.

Deferred from #653. (Distinct from the closed #337, which covered three pre-rework call sites.)

Issue #337, opened 2026-06-05
Refactor: consolidate WebP upload conversion into a single helper


svandragt:
Part of a small refactoring series (duplication / maintainability). No behaviour change intended.

## Problem

The "re-encode an upload to WebP, fall back to the original bytes on failure" control flow is written out three times, each wrapping the same primitives (`should_convert_to_webp()`, `convert_to_webp()`, `get_upload_dir()`):

- `src/response/upload.php:55-74` (`respond_upload()`)
- `src/micropub.php:500-518` (`saveUploadedPhotos()`)
- `src/micropub.php:787-795` (`respond_micropub_media()`)

The three copies differ only in how they obtain the source bytes (a `$_FILES` temp path, a PSR-7 `UploadedFileInterface` stream, another `$_FILES` temp path) and how they name the seed. The conversion decision, the `.webp` vs `.$ext` filename choice, and the fallback-to-original behaviour are identical and have already drifted slightly (e.g. `sha1($name)` vs `sha1($name . uniqid())`). A future change to conversion behaviour (quality, max dimension, adding a format) has to be made in three places.

## Suggested fix

Add one helper in `Lamb\Response` (next to the existing conversion primitives in `src/response/upload.php`) that takes a source path + client filename and returns the stored filename, doing the convert-or-fallback internally — e.g.:

```php
// returns the final filename written into get_upload_dir()
function store_upload_as_webp_or_original(string $srcPath, string $clientName, string $ext, ?string $seed = null): string
```

Then:
- `respond_upload()` passes the `$_FILES` temp path.
- `saveUploadedPhotos()` writes the PSR-7 stream to a tempfile and passes that.
- `respond_micropub_media()` passes its `$_FILES` temp path.

Keep `move_uploaded_file()` vs plain `rename`/`moveTo` semantics correct per caller (the media/upload endpoints have real uploaded files; the PSR-7 path is already a tempfile copy).

## Notes

- Follow red-green TDD: the conversion primitives are already unit-tested; add a test for the new wrapper covering (a) a JPEG/PNG is stored as `.webp`, and (b) a conversion failure falls back to storing the original extension.
- Pure refactor — `composer lint` + `composer analyse` must stay green and existing `MicropubAdapterTest` / upload tests must pass unchanged.

PR #683, opened 2026-08-22
Plan: converge the duplicate mechanisms behind the recurring bugs

svandragt:
Two documents, no code changes.

- **`PLAN.md`** — 11 PRs across 5 stages, each with its failing test, gates, size, risk and rollback.
- **`MODULARITY.md`** — the findings it cites, with file:line evidence for all ten duplication hotspots.

Lamb keeps producing the same kinds of bug because ten jobs in the codebase have two or three implementations each, and the copies drift. This plan collapses each of them to one.

## What has more than one way to do it

Ranked by likely bug production. Most are documented *by the code's own comments* — the divergence is a maintained convention, not an oversight.

1. **Saving a post — 9 entry points, 3 tiers of completeness.** Each independently decides whether to re-render, finalize the slug, notify subscribers. `wordpress.php:203` and `known.php:307` carry the same hand-maintained comment about not notifying. Underneath: **"published" is not an event** — `Webmention\process_outbound()` checks `is_publicly_visible()`, while `Websub\ping_scheduled_publishes()` infers publication from a `created > updated` watermark heuristic. Two definitions of one moment.
2. **5 HTTP transports**, one SSRF-guarded — and `Http\post_form()` has **zero callers**, while the two sites that would use it carry comments explaining why they don't.
3. **4 HTML sanitizers, 5 escapers** — including a *global* `escape()` declared inside `themes/base/feed.php` behind `function_exists()`, shadowing the namespaced helper.
4. **2 front-matter engines behind 4 assemblers**, 14 call sites across 8 files. Front matter *is* the post format.
5. **3 upload pipelines.** `response/upload.php:320-333` records the fallout: assets landed `0600` on one path and `0666 & ~umask` on the others, so a static-file server couldn't read images the site was serving.
6. **5 visibility expressions.** Docblocks record two already-shipped bugs from this.
7. **5 output serializers, 2 of them theme parts** — a theme owns the syndication contract.
8. **`2024/_items.php` and `2026/_items.php` differ by two lines, one a comment**, and both diverged from base. `_related.php:31` shows a fix hand-ported between copies.
9. 3 near-identical importer CLIs. 10. 6 config sources.

## Two live bugs found along the way

**1. Notify-after-failed-write, fixed on one save path only.** `redirect_edited()` (`response/posts.php:317-329`) carries this on its catch block:

> *"Falling through on a failed write … announced the unsaved content to webmention receivers and the WebSub hub."*

`redirect_created()` (`response/posts.php:48-58`) still calls `notify_post_subscribers($bean)` outside its try/catch. Narrower but live: `finalize_and_store_post()` stores twice, and if the *second* store throws (SQLite locked by a concurrent `/_cron` — the case that comment names) the bean already has an id, so neither `enqueue_for_post()` nor `ping_for_post()` early-returns. Webmentions get queued against an unstored slug and the hub is pinged for content that wasn't persisted. That's PR 1.

**2. `/_cron` can silently stop delivering webmentions.** `process_feeds()` (`network.php:102-149`) runs five jobs in one straight line with both notification drains **last**, then writes the run watermark. Nothing in the repo calls `set_time_limit()`, so `/_cron` inherits the host's PHP limit — typically 30s under the FPM setups the repo ships — while one feed fetch may take `FEED_FETCH_TIMEOUT` (15s). **Two slow feeds end the request inside the crawl loop**, skipping both drains *and* the watermark write; the next run is immediately due and walks into the same wall. One persistently slow feed stops outbound webmention delivery indefinitely, with no symptom but replies never arriving at their targets. The per-feed crawl is also unguarded — `network/ingest.php:116,141` catches `SQL` around individual stores, nothing wraps `crawl_feed()`. That's Stage 3.

## Plan shape

| Stage | PRs | What |
|---|---|---|
| 0 | 1 | The notify-after-failed-write twin |
| 1 | 2–7 | Convergence: dead code, one escaper, one front-matter engine, one upload store, theme-part dedupe, one `data_dir()` |
| 2 | 8 | `Post\save()` + lifecycle events — the D1 fix, largest change in the plan |
| 3 | 9 | `/_cron` ordering, per-feed guards, `finally`-scoped watermark, explicit time limit |
| 4 | 10 | Importer registry + one `bin/lamb import` driver (D9) |
| 5 | 11 | Atom/JSON Feed move from theme parts into code (D7), two-step deprecation |

Only one real dependency edge in the plan (PR 8 after PRs 4–5). Any subset is coherent on its own.

## Scope boundaries

`src/` stays flat — optional features are not moved into a module tree, and the 44% of `src/` that is optional features stays where it is. Lamb promises a plug-and-play install, so anything that would make a feature set something an operator selects is out of scope by design.

Nothing operator-visible: no new configuration, no toggles, no dependency changes. An install upgrading through all 11 PRs sees the same routes, the same features and the same `composer install`. The only intended operator-visible effect is a bug stopping happening. The one developer-visible change is PR 11, which is why it ships as a two-step deprecation rather than a cut — a theme overriding a feed part is someone's live blog.

## Verification

Documentation only — no code touched. Worth knowing for the plan's assumptions: PHPStan is at **level 8 with an empty baseline** and coverage is gated at **70%**. Both are load-bearing safety nets for the refactors proposed, and every stage's exit criteria assume they stay that way.

https://claude.ai/code/session_011ALrRmJaWTastLiW7WvDeb

### DIFF
--- a/MODULARITY.md
+++ b/MODULARITY.md
@@ -1,29 +1,15 @@
-# Modularity investigation — duplicate mechanisms and plugin seams
+# Duplicate mechanisms in Lamb
 
-**Status:** findings, nothing implemented. Written in response to "the project is
-sprawled, we keep finding bugs — find where there are multiple ways to do
-something and how we can split them off into plugins."
+Where the codebase does one job two or three different ways, with file:line evidence
+for each. This is the findings document behind [PLAN.md](PLAN.md), which sequences the
+work; it records what is duplicated, not what to do about it.
 
-**The sequenced plan derived from these findings is [PLAN.md](PLAN.md).** This
-document is the evidence base — what the plan cites — and is not itself a plan.
-
-**§1 and §2 stand. §3, §4 and §5 are superseded: the module system they propose was
-rejected.** The reason is in PLAN.md §Decision and is a product one, not a technical
-one — Lamb promises a plug-and-play install, and a `src/modules/` tree puts a disable
-switch ten lines away, which is the step toward plug-many-things-and-configure. The
-duplication findings in §2 are unaffected and are what the plan acts on; three items
-§3–4 justified as module plumbing survived as standalone bug-class fixes (the `/_cron`
-ordering, the importer registry, feeds moving out of themes). Read §3–5 as the
-rejected option, kept for the record.
-
-The finding in one line: **the sprawl and the bugs are two different problems, and
-only one of them is fixed by plugins.** The bugs come from ~10 places where the
-same job is done two or three different ways inside the core. The sprawl comes from
-~44% of `src/` being optional features fused into that core. Extracting a plugin
-does not fix a duplicate mechanism — it copies it into a module boundary and makes
-it harder to see. So the order matters: **converge first, extract second.**
-
----
+The through-line: **the bugs and the sprawl are separate problems.** The bugs come
+from the ten hotspots in §2, where the same job has more than one implementation and
+the copies have drifted. The sprawl is §1 — 44% of `src/` being optional features in a
+flat namespace. Fixing the duplication does not require touching the sprawl, and the
+reverse is not true: consolidating behind a boundary before the implementations agree
+just hides the divergence somewhere harder to see.
 
 ## 1. What "sprawled" measures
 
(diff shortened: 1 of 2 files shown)


svandragt:
Closing unmerged — the content moved to issues, which is the right home for it.

**Tracker: #684**, with 11 sub-issues.

Two reasons this shouldn't have been a PR:

1. **It partly duplicated tracked work.** #672 — *"Consolidate media-endpoint and /upload store-or-fallback into one helper"* — is the D5 finding, already filed. Its body also shows the same duplication has been attacked three times (#337 closed, #653, #672), which is better evidence for the finding than anything in the deleted doc. That collision was invisible from inside a markdown file.
2. **A plan file is transient.** Once the work lands, it describes a world that no longer exists and someone has to remember to delete it. Root-level docs are an enumerated set per the 2026-05-29 decision (`README`, `CONTRIBUTING`, `DECISIONS`, `DESIGN`, `PRODUCT`); a work plan isn't one.

Two live bugs were also sitting as prose in here, where nothing could close them. They're now #685 and #686.

| | Issue |
|---|---|
| **Bug** | #685 — create path notifies webmention/WebSub after a failed write |
| **Bug** | #686 — one slow feed can stop `/_cron` delivering webmentions indefinitely |
| D1 | #692 — post writes through one funnel; publication becomes an event |
| D2 | #687 — delete the unused `Http\post_form()` |
| D3 | #688 — one XML escaper; drop the global `escape()` in `themes/base/feed.php` |
| D4 | #689 — one front-matter engine behind one assembler |
| D5 | #672 — *(existing)* consolidate the upload store paths |
| D7 | #694 — Atom and JSON Feed move out of theme parts into code |
| D8 | #690 — dedupe `parts/_items.php` across base, 2024 and 2026 |
| D9 | #693 — importer registry and a single `bin/lamb import` driver |
| D10 | #691 — route the CLI importers through `Bootstrap\data_dir()` |

D6 (five expressions of "is this post visible") is recorded in #684 as a known hotspot with no scheduled fix — it is the most consolidated of the ten already, and its docblocks claim a single allow-list. Worth revisiting only if a visibility bug turns up.

Scope note carried into #684: `src/` stays flat. An earlier draft of this PR proposed moving optional features into a module tree; that is out of scope, because a module set an operator has to enable is the opposite of the plug-and-play install Lamb promises.

The branch is left in place if the drafts are worth referring back to; nothing from it merges.

---

Issue #717, opened 2026-08-22
Convert the remaining 8 write sites to Post\save()


svandragt:
Follow-up to #692, which landed the \`Post\save()\` funnel + event seam and converted the first site (\`apply_checkbox_toggle\`). Convert the remaining write sites one commit at a time behind the same green suite, until \`grep -rn 'notify_post_subscribers\|finalize_and_store_post' src/\` returns hits only inside \`src/post.php\` (the #692 exit criterion).

Order (each independently revertable, per #692):

- [ ] \`Response\redirect_created()\` — response/posts.php (finalize + notify)
- [ ] \`Response\redirect_edited()\` — response/posts.php (notify; lock + slug-change redirect — supplies the pre-edit slug that `save()`'s `redirect_on_slug_change` flag needs)
- [ ] \`Micropub::createCallback()\` — micropub.php (finalize + notify)
- [ ] \`Micropub::updateCallback()\` — micropub.php (notify)
- [ ] \`Network\create_item()\` — network/ingest.php (finalize, no notify)
- [ ] \`WordPress\import_item()\` — wordpress.php (finalize, no notify)
- [ ] \`Known\import_item()\` — known.php (finalize, no notify)
- [ ] \`Restore\import_post()\` — restore.php (finalize, no notify)

Each conversion: one test per site asserting which events fire with which context (the executable flag matrix). Also implement `redirect_on_slug_change` when converting the edit site. Refs #692, #684 (D1).

Issue #669, opened 2026-08-21
Make latest_content_timestamp() monotonic across trash/restore


svandragt:
`latest_content_timestamp()` computes `MAX(updated)` over live, published posts, so trashing the newest post moves the value backwards rather than only forward. A date-only client (no `If-None-Match`, e.g. some intermediary caches) can then be served a stale `304` for content that has since changed, because the `Last-Modified` it checks against dropped below what the cache holds.

Fixing this needs a stored high-water mark rather than a `MAX()` over live rows — and that mark must also survive restoring a post from trash.

Deferred out of scope from #667.

PR #728, opened 2026-08-23
Cover the media-endpoint upload store failure (#673)

### LINKED ISSUE #673: Add e2e coverage for a failed media-endpoint upload store
#653 fixed `respond_micropub_media()` returning 201 when `move_uploaded_file()` fails, but shipped without an automated test — reproducing the failure needs a real registered upload against an unwritable destination, which the unit suite cannot set up. The author verified manually (`chattr +i` on the destination) and offered a Playwright e2e case.

Add that test so the fix has regression coverage.

Follow-up from #653.

svandragt:
Closes #673. Adds the regression coverage #653's fix was missing, at two levels.

## Store layer (`tests/Unit/UploadTest.php`)
The existing test only used a non-convertible extension (gif), skipping the WebP branch. Adds a **convertible-extension** case (jpg bytes that aren't a real image) so both `store_webp_copy()` and the `move_uploaded_file()` fallback fail — asserting `store_upload_or_fallback()` returns `null` and writes neither file. jpg/png is the common real-world Micropub photo path.

## Endpoint (`tests/Unit/MicropubMediaEndpointTest.php`) — the actual #673 ask
Covers `respond_micropub_media()` mapping a failed store to **500, no Location**, not a phantom 201 (src/micropub.php:1620-1625).

This needed a small **DI seam**: `respond_micropub_media(?LambMicropubAdapter $adapter = null)`, defaulting to `new LambMicropubAdapter()` (router call site unchanged). Without it the function hardwired the adapter, forcing a real `fetch()` token-introspection call. Typed to the concrete `LambMicropubAdapter` (not the abstract base) to satisfy PHPStan level 8 — that's the class `StubMicropubAdapter` extends, so it's still fully injectable. The seam unblocks HTTP-entry Micropub testing generally.

The endpoint calls `exit()` on every path, which breaks `@runInSeparateProcess` result-passing, so the test drives it in a real `php` subprocess (proc_open) and reads back `http_response_code()` + the JSON error body. The 500 status and `server_error` body assertions are CLI/CI-safe (CI runs pcov/none, not Xdebug); the Location-absence check is an Xdebug-only bonus locally.

## Test plan
- [x] Red-green proven: forcing the null branch to fall through → RED (`201 !== 500`); restored → GREEN.
- [x] `vendor/bin/codecept run Unit`: 2106 tests green.
- [x] PHPStan level 8: no errors (baseline empty). `composer lint` clean.

Part of the 0.14.0 Convergence follow-up; base of a 3-PR stack (#673 → #669 → #717).

### DIFF
--- a/src/micropub.php
+++ b/src/micropub.php
@@ -1537,9 +1537,15 @@ function micropub_error(int $status, string $error, string $description, ?string
  * Handles Micropub media endpoint requests (POST multipart/form-data with a 'file' field).
  * Validates the bearer token, saves the uploaded file, and returns HTTP 201 + Location.
  *
+ * @param LambMicropubAdapter|null $adapter The adapter to verify the bearer token against;
+ *     defaults to a new LambMicropubAdapter. Injectable so tests can stub token
+ *     introspection instead of hitting the real token endpoint over HTTP. Typed to
+ *     the concrete class, not the MicropubAdapter base, because this function relies
+ *     on LambMicropubAdapter's narrower verifyAccessTokenCallback() return shape
+ *     (array{me, scope}|false) rather than the base's array|string|false|ResponseInterface.
  * @return void
  */
-function respond_micropub_media(): void
+function respond_micropub_media(?LambMicropubAdapter $adapter = null): void
 {
     $headers = getallheaders() ?: [];
     $authHeader = $headers['Authorization'] ?? $headers['authorization'] ?? '';
@@ -1563,7 +1569,7 @@ function respond_micropub_media(): void
         micropub_error(401, 'unauthorized', 'Missing bearer token.', bearer_challenge());
     }
 
-    $adapter = new LambMicropubAdapter();
+    $adapter ??= new LambMicropubAdapter();
     if (mp_debug_enabled()) {
         $adapter->logger = mp_adapter_logger();
     }
--- a/src/micropub.php
+++ b/src/micropub.php
@@ -1537,6 +1537,10 @@ function micropub_error(int $status, string $error, string $description, ?string
  * Handles Micropub media endpoint requests (POST multipart/form-data with a 'file' field).
  * Validates the bearer token, saves the uploaded file, and returns HTTP 201 + Location.
  *
+ * @param mixed $args The router's positional route arguments. Unused — this
+ *     endpoint reads $_FILES/$_POST directly — but declared first because
+ *     call_route() invokes every handler as $callback($args); a typed first
+ *     parameter would receive that array and fatal (a 500 on every request).
  * @param LambMicropubAdapter|null $adapter The adapter to verify the bearer token against;
  *     defaults to a new LambMicropubAdapter. Injectable so tests can stub token
  *     introspection instead of hitting the real token endpoint over HTTP. Typed to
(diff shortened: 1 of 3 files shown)


svandragt:
## Automated code review

**REQUEST CHANGES** — the PR as pushed breaks the media endpoint for every request and is the cause of the red CI.

### Blocking

- **`src/micropub.php:1547` — typed first parameter collides with the router's calling convention → HTTP 500 on every `/micropub-media` request.** `call_route()` (`src/routes.php:145`) invokes every handler as `$callback($args)`, so `respond_micropub_media` is called as `respond_micropub_media([])`. The old zero-param signature ignored that extra array; the new `respond_micropub_media(?LambMicropubAdapter $adapter = null)` binds `[]` to `$adapter`, which fails the type check with an uncaught `TypeError` and returns 500 before any auth logic runs. Verified with a minimal repro (`handler([])` against a `?Adapter $a = null` param throws `TypeError`).
  - **This is the CI failure.** `test (8.4)`/`test (8.5)` fail on `MicropubMediaCest:requestsWithNoBearerTokenAreRejected` (`tests/Acceptance/MicropubMediaCest.php:19`): *"Expected 401, Actual 500"*. The no-token path never reaches its `micropub_error(401, …)` because the `TypeError` fires on entry. Not an autoload/DB/php-binary/xdebug problem — it is the signature.
  - **Fix:** make the injected adapter a parameter the router never fills. Give the handler the router's `$args` as the first (ignored) param and the adapter second, e.g. `respond_micropub_media(mixed $args = null, ?LambMicropubAdapter $adapter = null)`, and call `respond_micropub_media(null, $adapter)` from the probe. (This exact fix already exists uncommitted in the working tree — it just needs to be committed and pushed; CI ran the old single-param commit `2131065`.)

### Non-blocking

- **`tests/Unit/MicropubMediaEndpointTest.php:143` — the no-Location assertion is a no-op under CI.** Headers are read via `xdebug_get_headers()` guarded by `function_exists(...)`; CI runs pcov/none, so in the subprocess that returns `[]` and the `foreach` over headers asserts nothing. The core `500` + `"error":"server_error"` assertions do hold without xdebug (`http_response_code()` works under CLI), and they already prove the `micropub_error(500)` early-exit path that precludes a Location header — so the header loop is redundant belt-and-braces rather than a coverage hole, but it gives a false sense of covering the header guarantee where it actually runs.
- **DI seam is otherwise sound.** In production the adapter stays `null` and the real `LambMicropubAdapter` is constructed via `??=`; token verification and the `has_micropub_scope(..., 'create')` check are unchanged. The injected adapter only substitutes token introspection, and `StubMicropubAdapter extends LambMicropubAdapter`, so the concrete type hint holds. No weakening of the auth path.
- **gif/store-failure setup is correct and deterministic.** GIF89a bytes sniff as `image/gif`, `should_convert_to_webp()` is false for gif, so it falls to `move_uploaded_file()`, which always refuses a non-HTTP-upload tmp file → `store_upload_or_fallback()` returns null → 500. The `UploadTest` additions cover the store-layer null path directly and cheaply; that is the higher-value, lower-cost half of the coverage.

svandragt:
Fixed in b437dd5: `respond_micropub_media()` now takes `(mixed $args = null, ?LambMicropubAdapter $adapter = null)` so the router's positional `$args` no longer binds to the typed adapter param. Acceptance `MicropubMediaCest` passes again (was the 500-not-401 CI failure); full Unit + Acceptance green locally.

PR #729, opened 2026-08-23
Back latest_content_timestamp() with a monotonic stored mark (#669)

### LINKED ISSUE #669: Make latest_content_timestamp() monotonic across trash/restore

svandragt:
Closes #669. Stacked on #728 (#673).

## Problem
`latest_content_timestamp()` computed `MAX(updated)` over live published posts, so trashing the newest post moved the value **backwards**. A date-only client/cache (no `If-None-Match`) could then be served a stale `304` for content that had since changed.

## Fix
Back it with a stored, monotonic high-water mark (`CONTENT_MODIFIED_TS` option) that only moves forward:
- `bump_content_timestamp()` sets the mark to `max(time(), previous+1)` (same-second-safe, mirroring `save_ini_text()`'s idiom), called at the three content-mutation chokepoints: `finalize_and_store_post()` (all create/edit), `soft_delete_post()`, `restore_post()`.
- `latest_content_timestamp()` reads the mark, folded with `config_modified_timestamp()` as before. **Lazy backfill**: when the mark is unset it initialises from the old `MAX(updated)` query, so existing installs don't reset to 0 on upgrade.

## Test plan (red-green)
- [x] `testTrashingNewestPostDoesNotLowerTimestamp` (core regression) — RED before fix (`2000 not >= 2030`), green after.
- [x] restore-doesn't-lower, backfill-from-existing-posts, and a deterministic bump-never-lowers-a-seeded-future-mark test (no wall-clock dependency).
- [x] `ConditionalRequestTest` (ETag/304) unaffected — 11 green.
- [x] Full Unit suite: 2110 green. PHPStan level 8 clean. `composer lint` clean.

Middle of the 3-PR stack (#673 → **#669** → #717).

### DIFF
--- a/src/micropub.php
+++ b/src/micropub.php
@@ -1537,9 +1537,15 @@ function micropub_error(int $status, string $error, string $description, ?string
  * Handles Micropub media endpoint requests (POST multipart/form-data with a 'file' field).
  * Validates the bearer token, saves the uploaded file, and returns HTTP 201 + Location.
  *
+ * @param LambMicropubAdapter|null $adapter The adapter to verify the bearer token against;
+ *     defaults to a new LambMicropubAdapter. Injectable so tests can stub token
+ *     introspection instead of hitting the real token endpoint over HTTP. Typed to
+ *     the concrete class, not the MicropubAdapter base, because this function relies
+ *     on LambMicropubAdapter's narrower verifyAccessTokenCallback() return shape
+ *     (array{me, scope}|false) rather than the base's array|string|false|ResponseInterface.
  * @return void
  */
-function respond_micropub_media(): void
+function respond_micropub_media(?LambMicropubAdapter $adapter = null): void
 {
     $headers = getallheaders() ?: [];
     $authHeader = $headers['Authorization'] ?? $headers['authorization'] ?? '';
@@ -1563,7 +1569,7 @@ function respond_micropub_media(): void
         micropub_error(401, 'unauthorized', 'Missing bearer token.', bearer_challenge());
     }
 
-    $adapter = new LambMicropubAdapter();
+    $adapter ??= new LambMicropubAdapter();
     if (mp_debug_enabled()) {
         $adapter->logger = mp_adapter_logger();
     }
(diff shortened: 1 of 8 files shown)


svandragt:
## Automated code review

**REQUEST CHANGES**

- Missing chokepoints: two live content-edit paths never call `bump_content_timestamp()`, so editing an existing post's content no longer advances `latest_content_timestamp()` once the mark is set. Before this PR, `MAX(updated)` picked these edits up automatically; now they're invisible to the validator, reintroducing stale-304 risk for edits — the same bug class this PR sets out to fix, via a different path.
  - `src/response/posts.php:335` — `redirect_edited()`, the admin edit-post form handler (reached from `respond_edit()`), stores the edited bean with a raw `R::store($bean)` and never bumps.
  - `src/micropub.php:567` — `MicropubAdapter::updateCallback()` (Micropub `update` action) also stores directly with `R::store($bean)`, no bump.
  - Fix: add `\Lamb\Response\bump_content_timestamp();` right after the successful `R::store($bean)` in both (same pattern as `soft_delete_post()`/`restore_post()`), or route both through the `Post\save()` funnel as part of the ongoing #717 conversion.

- No other blocking issues: `bump_content_timestamp()`'s `max(time(), previous+1)` is sound and same-second-safe; the lazy backfill correctly initialises from `MAX(updated)` on first unset read and is never re-derived afterward; `finalize_and_store_post()` and `save()` are confirmed separate store paths (`save()` does not call `finalize_and_store_post()`), so no double-bump between the two commits; the ETag/`send_304_if_current` path is untouched. Tests are appropriately scoped.

svandragt:
Fixed in 95dc075: added `bump_content_timestamp()` to both raw-store edit paths — `redirect_edited()` and Micropub `updateCallback()` — so an edit through either advances the monotonic mark (no stale 304). Added `testUpdateCallbackAdvancesTheContentTimestamp`. Full Unit (2112) + Acceptance (93) green. #717 later routes both through `save()`, which supersedes these explicit calls.

PR #753, opened 2026-08-26
Add scheduled image scan for base-image CVEs (#746)

### LINKED ISSUE #746: CI image-scan is non-blocking (exit-code 0), so new image CVEs only surface at release time
`ci.yml`'s `image-scan` runs Trivy with `exit-code: '0'` (report-only), while `release-artifacts.yml` runs it with `exit-code: '1'` (gates the push). So a newly-published base-image advisory passes CI green and only blocks at release — which is how 0.14.0's `kin-openapi` advisories stopped the Docker push after the release was already tagged (see #744).

Consider making CI's image-scan blocking (or a scheduled scan that opens an issue), so advisories are caught before a release is cut rather than during it.

Note the id-matching gotcha documented in #744: Trivy matches the exact id it prints and does not cross-map GHSA ↔ CVE, and which id it uses is DB/version-dependent — a runner upgrade can silently un-suppress an entry keyed by only one id.

svandragt:
Closes #746.

## What

A weekly (+ manual) workflow that rebuilds the release image and scans it with Trivy, then opens or auto-closes a single tracking issue when fixable HIGH/CRITICAL advisories appear.

## Why not the alternatives

- **Blocking CI's `image-scan`** would gate only `.docker/`-touching PRs — exactly the authors who can't fix a base-image OS/Go CVE, and vulnerable to the #744 id-matching flake turning a green PR red with no code change.
- **Dependabot** doesn't scan image contents, and the base image is already a floating tag (`dunglas/frankenphp:php8.4`), so there's no version bump to propose. It can't see this class of problem.

A scheduled scan catches a fresh advisory *before* a release is cut, without touching any PR.

## Behaviour

- Same scan rules as the release gate (`HIGH,CRITICAL`, `ignore-unfixed`, honouring `.trivyignore.yaml`).
- Findings → opens a tracking issue (or comments on the existing open one), labelled `security`, with the Trivy table.
- Clean run → closes the tracking issue if one is open.
- Never fails the run or gates a PR.

### DIFF
--- a/src/response/posts.php
+++ b/src/response/posts.php
@@ -18,7 +18,6 @@ use function Lamb\Post\populate_bean;
 use function Lamb\Post\sanitize_explicit_slug;
 use function Lamb\Post\save;
 use function Lamb\Post\toggle_rendered_checkbox;
-use function Lamb\Route\is_reserved_route;
 
 /**
  * Handles post creation from a form submission.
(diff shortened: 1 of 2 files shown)


svandragt:
Superseded and stale. The branch was crossed by concurrent sessions — its diff is actually the #740 reserved-slug change (already merged), not a scheduled scan. Direction also changed: #754 retires the published Docker image entirely, so no scheduled image scan is needed. Closing.

PR #777, opened 2026-08-29
Split sitemap.xml into a sitemap index past 50,000 URLs (#671)

### LINKED ISSUE #671: Split sitemap.xml into a sitemap index past 50,000 URLs
sitemaps.org caps a single sitemap document at 50,000 URLs. `sitemap_urls()` emits everything in one document regardless of count, so a large site produces an over-cap sitemap.

Split into a sitemap index once the count exceeds the cap. #664 fixed the sitemap's memory use but deferred this, treating it as a feature rather than a fix.

Deferred out of scope from #664.

svandragt:
## Summary
Closes #671. sitemaps.org caps a single sitemap document at 50,000 URLs, but `sitemap_urls()` emitted every visible URL in one `<urlset>`, so a large site produced an over-cap sitemap.

`/sitemap.xml` now adapts to the site size:
- **At or under 50,000 URLs:** a single `<urlset>`, byte-for-byte as before.
- **Over 50,000 URLs:** a `<sitemapindex>` listing child sitemaps at `/sitemap.xml/page/N`, and each `/sitemap.xml/page/N` serves that page's `<urlset>` slice.

The child URLs reuse the site's existing `/page/N` pagination convention, so no routing changes were needed — `extract_page_segment()` strips the segment and the handler reads `$_GET['page']`. A page request against an unsplit site, or a page past the last one, returns 404.

## Details
- `SITEMAP_MAX_URLS` (50,000) added to `constants.php`.
- `sitemap_urls()` gains optional `$page`/`$cap` params (defaults preserve the unpaginated behaviour); slicing is done with SQL `LIMIT`/`OFFSET`, keeping the id/slug/updated-only memory guard from #664.
- New `count_visible_posts()`, `sitemap_page_count()`, `render_sitemap_index()`.
- The cache key and path now include the page number, and the page count, so a visible-count change that crosses the split boundary (e.g. trashing or restoring an older post) mints a new generation rather than serving a stale index or urlset. Generation-based pruning keeps sibling pages of the current generation.

## Test plan
- 19 new unit tests in `SitemapTest.php`: page-count math, `sitemap_urls()` slicing and boundaries (exactly the cap, cap+1, last partial page), `render_sitemap_index()` structure/escaping/lastmod, per-page cache paths, generation-based pruning, and the split↔unsplit transition (cache key and served document shape track the visible count, not just the newest `updated`).
- `composer lint`, `composer analyse` (PHPStan level 8): clean.
- `vendor/bin/codecept run Unit`: 2165 tests green. `vendor/bin/codecept run Acceptance`: 96 tests green.

## Docs
- `docs/search-engines.md` and `src/response/README.md` updated to document the index split.

### DIFF
--- a/docs/search-engines.md
+++ b/docs/search-engines.md
@@ -36,6 +36,15 @@ Unlike the timeline, the sitemap **does** include
 The sitemap is cached and supports conditional requests, so crawlers that
 revisit it only re-download it when your content has actually changed.
 
+### Large sites
+
+[sitemaps.org](https://www.sitemaps.org/) caps a single sitemap document at
+50,000 URLs. Once your site passes that, `/sitemap.xml` automatically becomes
+a sitemap index instead of a plain sitemap: it lists child sitemaps at
+`/sitemap.xml/page/1`, `/sitemap.xml/page/2`, and so on, each holding up to
+50,000 URLs. Crawlers that follow sitemap indexes (all major ones do) pick
+this up with no configuration change on your part.
+
 ## robots.txt
 
 `/robots.txt` allows crawling, advertises the sitemap, and asks crawlers not to
--- a/src/constants.php
+++ b/src/constants.php
@@ -54,3 +54,7 @@ define('SESSION_LOGIN', 'logged_in');
 define('SUBMIT_CREATE', 'Create post');
 define('SUBMIT_EDIT', 'Update post');
 define('SUBMIT_LOGIN', 'Log in');
+// sitemaps.org caps a single sitemap document at 50,000 URLs. /sitemap.xml
+// splits into a sitemap index of /sitemap.xml/page/N children once the site's
+// visible URL count passes this.
+define('SITEMAP_MAX_URLS', 50000);
--- a/src/response/README.md
+++ b/src/response/README.md
@@ -200,11 +200,27 @@ The sitemap is served from disk. `respond_sitemap()` reads the validator
 first — one indexed row for the newest visible `updated`, plus the config
 timestamp — so a crawler revalidating a copy it already holds is answered 304
 without the document being built at all. Those same two values, together with
-`ROOT_URL`, key a cached copy under `data/cache/sitemap-<key>.xml`, so a
+`ROOT_URL`, key a cached copy under `data/cache/sitemap-<key>-<page>.xml`, so a
 request that does get a 200 streams a file instead of rebuilding 25,000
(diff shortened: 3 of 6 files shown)


svandragt:
Applied a follow-up fix from a code review (commit on this branch):

The disk cache key correctly folds in `page_count`, but the 304/ETag layer (`feed_cache($updated)`) keyed only on the newest post. So a shape change — `sitemapindex`↔`urlset`, or a change in child-page count — that leaves the newest post untouched (a non-newest post trashed, or a scheduled post crossing into visibility) was masked by a 304: a crawler kept serving itself a stale index pointing at a now-404 child page.

`feed_cache()` gains an optional shape discriminator, folded into the ETag's content timestamp; the sitemap passes its `page_count`. `$page` itself is part of the request URL (conditional GET is per-URL, and a shrunk count 404s before `feed_cache`), so `page_count` alone closes the gap. Added a red-green unit test asserting the emitted ETag turns over when the shape changes. Full Unit suite green (2166), phpcs + phpstan clean.

svandragt:
## Automated code review

**APPROVE**

Reviewed the sitemap-index split (`src/response/discovery.php`, `feeds.php`, `constants.php`) against current `main` (no merge conflicts; CI green on all 9 checks).

- Verified the `sitemap_urls()` page/offset arithmetic (home entry at index 0, page 1 carrying `cap-1` posts, page *k>1* offset `(k-1)*cap-1`) against the new `SitemapTest.php` cases, including the last-partial-page and split/unsplit transition tests — the math checks out.
- Checked whether an out-of-range or negative `?page=` could bypass the 50,000-URL cap and force an unbounded, uncached-forever full-table render via `respond_sitemap()`'s page-slice branch. It can't: `index.php`'s existing legacy-`?page=` redirect (lines 67-82) intercepts any `page` query param first and clamps it to `max(1, …)` before the route ever dispatches, and `Http\extract_page_segment()`'s `/page/(\d+)/` regex only ever yields non-negative integers. So the "else" branch in `respond_sitemap()` only ever sees `$page >= 1`.
- `sitemap_cache_key()`/`feed_cache()` folding in `page_count` as a shape discriminator correctly closes the stale-304 gap the PR describes (a non-newest post trashed/restored crossing the split boundary).

No blocking issues. Test suite not re-run locally (relying on this PR's own green CI); reviewed the diff and test additions directly instead.

---

PR #778, opened 2026-08-29
Add unit coverage for the /_cron run: extract run_feed_cycle() (#706)

### LINKED ISSUE #706: Add integration coverage for /_cron: feed ingestion and drain resilience
Follow-up from #686. `process_feeds()` (the `/_cron` handler) is not unit-reachable — it calls `header()`, acquires a file lock, and ends in `exit()`/`die()` — so #686 's guarantees are only partly covered:

- **Unit:** `crawl_feed_guarded()` (one bad feed → error shape, no propagation).
- **Manual only:** that a slow/failing feed does not abort the run and that the drains + watermark still run. Verified once by hand with two local Lamb installs (one serving a feed, one consuming) plus a deliberately slow feed shown to fail at `FEED_FETCH_TIMEOUT` while the run still reached `Done` and advanced the watermark.

Nothing automated pins the orchestration, so it can regress silently.

## Suggested
A Playwright/acceptance test (or a small refactor extracting the loop+drains into a testable unit) that asserts, against a running instance: a failing feed yields a `FAILED` line, a good feed alongside it still ingests, the response reaches `Done`, and the watermark advances.

Note: the SSRF guard blocks loopback/private feed hosts, so a local harness needs an opt-in allowance hook (the manual test used a temporary env-gated one).

svandragt:
## Summary
Closes #706. `process_feeds()` (the `/_cron` handler) is not unit-reachable — it sends a header, takes a file lock and ends in `exit()`/`die()` — so the run's orchestration was covered only by a manual two-install test and could regress silently. Only `crawl_feed_guarded()` had unit coverage.

This extracts the crawl loop plus the drains-in-a-`finally` and the watermark advance into a new `run_feed_cycle()`, with every collaborator injectable (feeds, crawler, both drains, watermark setter, output sink, clock). `process_feeds()` keeps the header, lock, rate gate, maintenance lines and `exit('Done')`, and calls `run_feed_cycle()` for the crawl + drain body. Real `/_cron` requests are unchanged — the defaults are the production collaborators and the output is byte-for-byte the same.

I chose the refactor-to-testable-unit path the issue offers rather than a Playwright/acceptance test: the acceptance route needs an SSRF opt-in for a loopback fixture host, and adding a private-host bypass to the fail-closed guard is not worth it for a test. The unit approach pins the same orchestration properties with no network and no security surface.

## Test plan
New `tests/Unit/RunFeedCycleTest.php` drives `run_feed_cycle()` with injected closures — no network — and pins:
- a failing feed produces a `FAILED` line;
- a good feed alongside a failing one still ingests;
- the run reaches the drains even when a feed fails;
- the watermark setter receives the advanced timestamp;
- both drains run even when the crawl loop throws (and the throw still propagates);
- a not-yet-due feed is skipped without calling the crawler.

- `vendor/bin/codecept run Unit`: 2155 tests green (+6 new), 2 pre-existing skips, no failures.
- `phpcs`: clean. `phpstan` (level 8): no errors.

### DIFF
--- a/src/network.php
+++ b/src/network.php
@@ -99,6 +99,79 @@ function webmention_line(array $sent): string
     );
 }
 
+/**
+ * Crawls every due feed, then drains the notification queues and advances the
+ * run watermark.
+ *
+ * Extracted from process_feeds() so the run's orchestration is unit-testable —
+ * process_feeds() itself is not, because it sends a header, takes a file lock
+ * and ends in exit()/die(). Every collaborator is injectable so a test drives
+ * the loop and drains without touching the network; real /_cron requests get
+ * the production collaborators from the defaults, and the output is byte-for-byte
+ * what process_feeds() emitted inline before.
+ *
+ * The watermark advance and both drains run in a `finally`, so a feed that fails
+ * — or the loop throwing outright — can never starve webmention/WebSub delivery.
+ * The watermark is advanced before draining so a partial run still rate-limits
+ * the next one even if a drain throws; the drains then retry on the next run.
+ *
+ * @param array<string, string>|null $feeds Feed name => URL map (default get_feeds()).
+ * @param callable(string, string): array{ok: bool, items: int, error: ?string}|null $crawler
+ *        Per-feed crawler (default crawl_feed_guarded(), which never throws).
+ * @param callable(): int|null $websub_drain WebSub drain returning the ping count
+ *        (default ping_scheduled_publishes()).
+ * @param callable(): array{sent: int, failed: int, skipped: int, cancelled: int}|null $webmention_drain
+ *        Outbound-webmention drain (default process_outbound()).
+ * @param callable(int): void|null $advance_watermark Persists the run timestamp
+ *        (default writes the last_processed_date option).
+ * @param callable(string): void|null $output Output sink (default echo).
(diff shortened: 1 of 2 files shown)


svandragt:
## Automated code review

**APPROVE**

Reviewed the `run_feed_cycle()` extraction from `process_feeds()` (`src/network.php`). No merge conflicts against current `main` (verified via `git merge-tree`; GitHub's own `mergeable_state` was still "unknown" at review time), CI green.

- Diffed the extracted body against the original inline `try/finally` block line-by-line: the watermark-advance-before-drains ordering, the "drains always run even on a feed failure or an uncaught throw" guarantee, and the per-feed `feed_fetch_due()` skip are all preserved exactly, just parameterized behind defaults that reproduce the original collaborators (`crawl_feed_guarded()`, `ping_scheduled_publishes()`, `process_outbound()`, `time()`).
- `process_feeds()`'s `advance_watermark` closure correctly reuses the already-fetched `$cron_last_updated` option bean rather than re-reading it, matching the PR description.
- New `RunFeedCycleTest.php` covers the orchestration properties that actually matter here (failure isolation, drain-always-runs, watermark value, skip-when-not-due) with no network access.

No blocking issues.

---

PR #780, opened 2026-08-30
Force busy_timeout=0 for the login-throttle reservation's write lock

svandragt:
## Bug

`reserve_login_attempt()` (#766) documents and tests a fail-closed contention path: when another request already holds the `option` table's write lock, `BEGIN IMMEDIATE` is supposed to throw immediately so the attempt is refused rather than let through. That test (`testReserveLoginAttemptRefusesRatherThanThrowingWhenTheWriteLockIsHeld`) only reproduces this by manually forcing `PRAGMA busy_timeout = 0` on both connections — production code never does this.

A bare `new PDO('sqlite:...')` (which is all `bootstrap_db()` ever creates) keeps whatever busy_timeout the host's SQLite build defaults to — 60000ms (60s) on this container, confirmed with:

```php
$pdo = new PDO("sqlite:...");
$pdo->query('PRAGMA busy_timeout')->fetchColumn(); // => 60000
```

With that default, a contended `BEGIN IMMEDIATE` doesn't throw — it silently *blocks* until the lock frees or 60s passes. That defeats the design intent: the request just queues instead of being refused, still runs `password_verify()` once its turn comes (the exact bcrypt pile-up #742/#766 exist to prevent, just serialised across the stall instead of parallel), and ties up a web-server worker/connection for the wait. `prune_login_throttle()`'s per-row `R::trash()` calls, which run before the `BEGIN IMMEDIATE`, were also unguarded against this contention.

## Verified over real HTTP

Started `php -S 127.0.0.1:8747 -t src`, held the SQLite write lock on `data/lamb.db` from a second, independent OS process (`BEGIN IMMEDIATE` + sleep), then fired a real concurrent wrong-password `POST /login` at the live server:

- **Before the fix:** the request blocked for the full 6s lock hold, then proceeded normally past the throttle and ran `password_verify()` — `200` "Password is incorrect" after ~5.8s.
- **After the fix:** the request refuses in ~5ms — `429` "Too many failed attempts. Try again in 1 minute."

## Fix

Forces `busy_timeout` to `0` around the reservation's critical section (prune + `BEGIN IMMEDIATE` + read/write/`COMMIT`) in `reserve_login_attempt()`, restoring the connection's previous value in a `finally` block so the rest of the request (and every other route) keeps its normal, forgiving timeout. Also rolls back a partially-opened transaction before returning the refusal, since a caught busy error can now genuinely leave one open.

## Tests

- Added `testReserveLoginAttemptRefusesPromptlyUnderContentionWithoutTestForcedBusyTimeout` to `tests/Unit/LoginThrottleTest.php`: unlike the existing contention test, it does **not** force `busy_timeout` on the RedBean side (reproducing real production conditions), and holds the write lock from a genuine separate OS process (`Symfony\Component\Process\Process`, already a dev dependency) so a regression fails in bounded time (the hold's length) instead of hanging the suite. Confirmed **red** against the pre-fix code (assertion failure: `retry_after` was `0`, and it took the full 2s hold instead of refusing promptly) and **green** after.
- Full suite: `vendor/bin/codecept run Unit` — 2150 tests (2149 + 1 new), 0 failures, 2 skipped (unchanged from baseline).
- `vendor/bin/codecept run Acceptance` — 96 tests, 0 failures (includes `LoginCest`).
- `vendor/bin/phpstan analyse` — no errors. `composer lint` (`phpcs`) — clean.

## Update: WAL fix for a reader-vs-COMMIT regression

Review of this branch found that `busy_timeout=0` was reasoned about only for the `BEGIN IMMEDIATE` writer-vs-writer case. But the reservation's `COMMIT` upgrades to an EXCLUSIVE lock, which SQLite cannot take while any other connection holds a read lock. The DB runs in the default rollback journal (no WAL), so with `busy_timeout=0` that `COMMIT` returns `SQLITE_BUSY` immediately against an ordinary concurrent reader — a second visitor merely rendering a page could make a correct-password login refuse with the throttle message.

Fixed by enabling WAL at setup (`PRAGMA journal_mode=WAL` in `bootstrap_db()`): a writer commits without waiting for readers, so `busy_timeout=0` now refuses only on a genuine concurrent writer, which is the intent. Added `testBootstrapDbEnablesWalSoLoginCommitsSurviveConcurrentReaders` (runs `bootstrap_db()` in a subprocess, asserts the file's `journal_mode` is `wal`).

Full suite re-run on the fix: `vendor/bin/codecept run Unit` — 2151 tests, 0 failures (2 pre-existing skips); `vendor/bin/codecept run Acceptance` — 96 tests, 0 failures. `composer lint` and `composer analyse` (level 8) clean.

### DIFF
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -102,6 +102,15 @@ function bootstrap_db(string $data_dir): void
         }
     }
     R::setup(sprintf("sqlite:%s/lamb.db", $data_dir));
+    // WAL lets a writer commit without waiting for concurrent readers. The login
+    // throttle's reserve_login_attempt() forces busy_timeout=0 so a contended
+    // reservation fails closed instead of stalling; without WAL its COMMIT takes
+    // an EXCLUSIVE lock that any page-render read holds off, so an ordinary
+    // concurrent visitor would make a correct-password login refuse. WAL narrows
+    // that refusal to a genuine concurrent writer, which is the intent.
+    // ponytail: WAL needs a local filesystem; self-hosted installs on NFS would
+    // fall back to the slower rollback journal and lose this guarantee.
+    R::exec('PRAGMA journal_mode=WAL');
     R::useWriterCache(true);
 
     ensure_post_columns();
--- a/src/response/README.md
+++ b/src/response/README.md
@@ -99,6 +99,18 @@ retrying cannot extend a block. `client_ip()` reads `REMOTE_ADDR` only; see
 [AGENTS.md](../../AGENTS.md)'s Security section for why `X-Forwarded-For` is
 not trusted here.
 
+The peek only rejects an already-blocked client; the race-free gate is
+`reserve_login_attempt()`, which takes a `BEGIN IMMEDIATE` write lock and
+checks-and-increments the counter atomically before bcrypt. Without it a
+concurrent burst from one IP could all read the same under-limit count and each
+run `password_verify()`, serialising only on the write — the bcrypt pile-up the
+throttle exists to stop. SQLite's busy handler is left at the host default
+(tens of seconds), so the reservation forces `busy_timeout = 0` for that
+critical section: a contended lock then refuses immediately (the client retries)
+instead of stalling every attempt behind the lock and reopening the pile-up. The
+lazy prune runs inside that same zero-timeout window on purpose — a prune that
+blocked on contention would reopen it just as surely.
+
 Post-login, `local_redirect_target()` constrains `?redirect_to=` to a local
 path so the parameter can't be turned into an open redirect — the same
 protocol-relative-path check reappears in `posts.php` (see below), because a
--- a/src/response/auth.php
+++ b/src/response/auth.php
@@ -376,37 +376,71 @@ function login_throttle_retry_after(string $ip, int $now): int
  * RedBeanPHP\Facade::begin()), so the lock is taken with a raw statement
  * instead.
  *
+ * Neither PHP nor RedBeanPHP ever configures SQLite's busy handler, so the
+ * connection keeps whatever a bare `new PDO('sqlite:...')` defaults to on the
+ * host's SQLite build — commonly tens of seconds, not 0. Left alone, a
+ * contended BEGIN IMMEDIATE below would silently *block* for that long
+ * instead of throwing: every concurrent attempt in a burst would still queue
+ * up and eventually reach password_verify() once its turn came, exactly the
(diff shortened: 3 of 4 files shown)


svandragt:
Reviewed a code-review finding against this PR and it turned out to be a false positive worth recording:

The finding read the docblock — which said prune keeps the host's normal, forgiving timeout — and concluded that running `prune_login_throttle()` inside the `busy_timeout=0` window was an accidental placement that could refuse legitimate logins.

But `testReserveLoginAttemptRefusesPromptlyUnderContentionWithoutTestForcedBusyTimeout` encodes the opposite intent: a prune that blocked on contention would stall behind the lock and reopen the exact bcrypt pile-up this PR closes. Moving prune out (as the finding suggested) makes that test fail. So the **code is correct** — the **docblock was stale**.

Pushed a docs-only commit: corrected the docblock's claim, trimmed it (it was ~42 lines), and moved the deep rationale into `response/README.md`'s throttle section. No behaviour change.

(Heads-up: that contention test passes in CI but fails deterministically on my local machine — a `php` subprocess/timing issue unrelated to this change; I'll park it separately.)

svandragt:
## Automated code review

**APPROVE**

Gave this one extra scrutiny as auth/throttle code (`src/response/auth.php`'s `reserve_login_attempt()`). No merge conflicts against current `main`, CI green.

- Traced the fail-closed contention path end to end: `busy_timeout` is forced to `0` before `prune_login_throttle()` + `BEGIN IMMEDIATE` and unconditionally restored via the previous value read up front, in a `finally`, even on an uncaught path. A `SQL` exception at any point in the critical section (including during prune) now rolls back an open transaction (`$in_transaction` correctly tracks whether `BEGIN IMMEDIATE` actually succeeded) and refuses the attempt rather than letting it through unreserved.
- The one real behavior change — a *prune* hitting lock contention now refuses a legitimate login rather than proceeding — is intentional and documented (`response/README.md`), and matches the new `testReserveLoginAttemptRefusesPromptlyUnderContentionWithoutTestForcedBusyTimeout` test's intent. I agree with the author's own resolution of the earlier "move prune out" review finding as a false positive (see PR comment): moving it would reopen the exact bcrypt pile-up this PR closes.
- The subprocess-based test (genuine separate OS process holding the write lock, no test-forced `busy_timeout`) is a meaningfully stronger reproduction of production conditions than the existing same-process test.
- `PRAGMA busy_timeout = ' . $previous_busy_timeout` concatenates a value read back as `(int)` from SQLite, so no injection risk despite the string-built PRAGMA.

No blocking issues.

---

PR #786, opened 2026-08-31
WXR import: drop numeric slugs even without a redirectable old path

svandragt:
## Bug

`wordpress_status_path()` (`src/wordpress.php`) handles one WordPress export shape: a title-less post whose `<wp:post_name>` is purely numeric. WordPress puts the post id in the slug column when there's no title to build a slug from. Pinning that slug locally would produce a Lamb URL like `/26`, which collides with the canonical `/status/<id>` shape and can shadow it on lookup. So the importer should always drop it and fall through to `/status/<id>`.

The function conflated two separate questions:

1. Is `<wp:post_name>` purely numeric? This is what should gate slug-dropping.
2. Can we also derive a redirect path from `<link>`?

It returned `null` when either was false, and `import_item()` read that single result to decide whether to drop the slug:

```php
$slug = wordpress_status_path($item) !== null ? '' : (string) ($item['slug'] ?? '');
```

## Why it's reachable

A WordPress site on the default Plain permalink structure exports `<link>` as a bare query string, e.g. `<link>https://oldsite.example/?p=59</link>` — no path. WordPress still records the numeric `post_name` internally. So a title-less post from such a site is exactly the documented case (numeric slug `59`), but `wordpress_status_path()` returns `null` because `parse_url('?p=59', PHP_URL_PATH)` yields no path — not because the slug isn't numeric. `import_item()` then reads `null` as "not the numeric case" and keeps the slug, reproducing the `/59` vs `/status/<id>` shadowing the function exists to prevent. Plain permalinks are the out-of-the-box WordPress default, so this is common, not an edge case.

## Fix

Split the numeric-slug detection into its own function, `is_numeric_wordpress_slug()`, and call it directly from `import_item()` for the slug-dropping decision. `wordpress_status_path()` now delegates to it for the numeric test and keeps its original job: deriving an old path to redirect from, which can legitimately be `null` when `<link>` has no usable path.

No other behaviour changes: existing redirect tests for pretty-permalink sources (`/status/59/`, `/bugs/26/`) still pass unchanged.

## Verification

- Standalone reproduction script (not committed) calling `Lamb\WordPress\import_item()` directly with an item shaped like the scenario above (`link` = `guid` = `https://oldsite.example/?p=59`, `slug` = `'59'`). Before the fix: `bean->slug` was `'59'`. After: `bean->slug` is `''`.
- Regression test added to `tests/Unit/WordPressImportTest.php` (`testImportItemDropsNumericSlugEvenWhenLinkHasNoUsablePath`), asserting the slug is dropped, no `slug:` line is written to the body, and no redirect row is stored (there's no old path to redirect from).
- Confirmed red-before/green-after against `git show HEAD:src/wordpress.php` (fails: expected `''`, got `'59'`) and against the fix (passes).
- Full suite: `vendor/bin/codecept run Unit` — 2150 tests / 3957 assertions, 0 failures (up from 2149/3953 on `main`, the one new test).
- `vendor/bin/phpstan analyse` — no errors.
- `composer lint` (`phpcs`) — clean.

### DIFF
--- a/src/wordpress.php
+++ b/src/wordpress.php
@@ -230,7 +230,7 @@ function import_item(array $item, callable $downloader, bool $dry_run = false, ?
         : prepare_imported_html($body_html, (string) $item['created'], $downloader);
     $markdown = html_to_markdown($prepared);
     $tags = array_values(array_map(static fn($t): string => (string) $t, (array) ($item['tags'] ?? [])));
-    $slug = wordpress_status_path($item) !== null ? '' : (string) ($item['slug'] ?? '');
+    $slug = is_numeric_wordpress_slug($item) ? '' : (string) ($item['slug'] ?? '');
     $body = build_post_body((string) $item['title'], $markdown, $tags, $slug);
 
     $bean = populate_bean($body, null, null, $bean);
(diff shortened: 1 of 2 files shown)


svandragt:
## Automated code review

**APPROVE**

Reviewed the WXR numeric-slug fix in `src/wordpress.php`. No merge conflicts against current `main`, CI green.

- Confirmed the two concerns are now correctly decoupled: `import_item()` (`src/wordpress.php:233`) gates slug-dropping on the new `is_numeric_wordpress_slug()` alone, while `wordpress_status_path()` (`src/wordpress.php:288`, used only for redirect creation) is untouched at its call site and still independently requires a usable `<link>` path. So a "Plain"-permalinks source (`?p=59`, no path) now drops the numeric slug but — correctly — still stores no redirect, since there's genuinely no old path to redirect from.
- The new regression test (`WordPressImportTest::testImportItemDropsNumericSlugEvenWhenLinkHasNoUsablePath`) exercises exactly this combination and asserts both halves (slug dropped, no redirect row).
- No behavior change for the pretty-permalink case (existing tests unaffected, per the PR description).

No blocking issues.

---

Issue #731, opened 2026-08-23
Deciding a post is visible — converge the five expressions (D6)


svandragt:
Sub-issue of #684 (hotspot **D6**).

Deciding whether a post is visible has **five expressions** that drift:

- `visible_clause()`
- `public_posts_clause()`
- `is_viewable()` / `is_publicly_visible()` / `is_scheduled()`
- `websub.php:157`, re-assembling `SQL_PUBLISHED` inline
- an `is_menu_item()` backstop inside each theme's `_items.php`

The docblocks around these already record two shipped bugs caused by the copies diverging. Converge them to one definition of "visible", with the SQL clause and the PHP predicate derived from a single source.

## Expectations

Per #684 and AGENTS.md: red-green TDD, characterisation test first so "no behaviour change" is enforced rather than asserted. Gates stay green: `composer lint`, `composer analyse` (level 8, empty baseline), `vendor/bin/codecept run`, `composer coverage` (≥70%), `pnpm test`, `composer audit`.

Anything addressed by name keeps a deprecated shim for one release.

PR #788, opened 2026-09-01
Delegate the Restore/Delete-link deleted check to is_deleted()

svandragt:
## Bug

Three call sites in the theme layer decide whether to show **Restore** or **Delete** for a post using the raw check `$bean->deleted` / `$post->deleted`, instead of the canonical `is_deleted()` predicate (`($post->deleted ?? null) == 1`) used everywhere else a post's deleted state matters (`is_viewable()`, `is_publicly_visible()`, `preview_token_valid()`):

- `render_post_list()` in `src/theme.php` — shared by the 2024/2026 themes' `_items.php`, extracted in #731
- `src/themes/base/parts/_items.php` — the base theme's own copy of the same ternary
- `src/themes/base/parts/edit.php` — `if (!$post->deleted)` gating the Delete link

These render on every logged-in post list, status page, and edit page — hot paths, not edge cases.

No live bug today. Every write path stores `0`, `1`, or `null` in the `deleted` column (`soft_delete_post()`/`restore_post()` in `src/response/posts.php`), so the raw check and `is_deleted()` agree on every value actually written. But a non-canonical value like `deleted = 2` would make them disagree: the raw check shows **Restore** (implying the post is trashed) while `is_deleted()` says it is not, and `edit.php` hides **Delete** for a post that still needs it. Same drift as #749, #751, and #779.

## Fix

Replace the raw check with `is_deleted($bean)` at all three sites, adding `use function Lamb\is_deleted;` to the two theme parts that lacked it. `theme.php` uses the existing fully-qualified `\Lamb\is_deleted(...)` style already established there.

## Verification

- Standalone script requiring the real `src/` files, showing the divergence for `deleted = 2`: the raw check says "deleted", `is_deleted()` says "not deleted".
- A second script rendered all three real call sites for a `deleted = 2` post — red before the fix (Restore shown / Delete hidden), green after (all three agree with `is_deleted()`).
- Regression tests pinning both the canonical (`deleted = 1`) and non-canonical (`deleted = 2`) cases:
  - `tests/Unit/ThemeModernItemsPartTest.php` (`render_post_list()`, both themes)
  - `tests/Unit/ThemeItemsEscapeTest.php` (base theme's `_items.php`)
  - new `tests/Unit/ThemeEditPartTest.php` (base theme's `edit.php`)
- Ran the new and changed test methods against this branch's real source — all pass. The full pre-existing `ThemeModernItemsPartTest.php` suite, including its byte-for-byte HTML fixture pins, is unchanged and still passing, confirming no behaviour change for canonical values.
- `php -l` clean on all six touched files.
- `vendor/bin/phpcs` (the `composer lint` command) clean across the repo.
- `vendor/bin/phpstan analyse` (level 8) clean across `src/`.

**Environment note:** this session ran in an isolated git worktree with no local `vendor/`. The shared checkout's `vendor/autoload.php` hardcodes absolute paths back to the shared checkout's own `src/`, so a plain `vendor/bin/codecept` run would test the unmodified shared files, not this branch's. The red/green scripts and the new test methods were run through a custom bootstrap that loads this worktree's own `src/` instead; `phpcs`/`phpstan` took this worktree's paths directly. Run the full `vendor/bin/codecept run Unit` in CI against the pushed branch to confirm no regression the reflection-based spot run missed.

### DIFF
--- a/src/theme.php
+++ b/src/theme.php
@@ -497,7 +497,7 @@ function render_post_list(bool $hide_author): void
             <?= syndication_links($bean) ?>
 
             <?php if (isset($_SESSION[SESSION_LOGIN])) : ?>
-                <small><?= link_source($bean) ?> <?= action_preview($bean) ?> <?= action_edit($bean) ?> <?= $bean->deleted ? action_restore($bean) : action_delete($bean) ?></small>
+                <small><?= link_source($bean) ?> <?= action_preview($bean) ?> <?= action_edit($bean) ?> <?= \Lamb\is_deleted($bean) ? action_restore($bean) : action_delete($bean) ?></small>
             <?php endif; ?>
         </article>
         <?php // phpcs:ignore Generic.WhiteSpace.ScopeIndent.Incorrect -- leading whitespace here is literal output, preserved from the pre-extraction template
--- a/src/themes/base/parts/_items.php
+++ b/src/themes/base/parts/_items.php
@@ -3,6 +3,7 @@
 global $data;
 global $template;
 
+use function Lamb\is_deleted;
 use function Lamb\Theme\action_delete;
 use function Lamb\Theme\action_edit;
 use function Lamb\Theme\action_preview;
@@ -40,7 +41,7 @@ else :
             <div class="e-content"><?= anchor_headings($bean->transformed, !empty($bean->title) ? 3 : 2) ?></div>
             <?= syndication_links($bean) ?>
             <footer>
-                <small><?= action_preview($bean) ?> <?= action_edit($bean) ?> <?= $bean->deleted ? action_restore($bean) : action_delete($bean) ?></small>
+                <small><?= action_preview($bean) ?> <?= action_edit($bean) ?> <?= is_deleted($bean) ? action_restore($bean) : action_delete($bean) ?></small>
             </footer>
         </article>
         <?php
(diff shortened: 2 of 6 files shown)


svandragt:
## Automated code review

**APPROVE**

Reviewed the `is_deleted()` consolidation across `src/theme.php`, `src/themes/base/parts/_items.php` and `src/themes/base/parts/edit.php`. No merge conflicts against current `main`, CI green.

- Confirmed `is_deleted()`'s definition (`src/lamb.php:878-881`, `($post->deleted ?? null) == 1`) matches the PR's description and is strictly narrower than the raw truthy checks it replaces, same shape as the prior `is_deleted()`/`is_unpublished()`/`is_draft()` consolidations (#749/#751/#779).
- Checked out the branch and ran it for real (this session's `vendor/` is actually populated despite the standing task note saying otherwise): `vendor/bin/phpstan analyse` clean, and the three new/changed test files (`ThemeEditPartTest`, `ThemeItemsEscapeTest`, `ThemeModernItemsPartTest`) all pass — 19 tests, 30 assertions. Full `vendor/bin/codecept run Unit` — 2156 tests, 3964 assertions, 0 failures (2 pre-existing skips), matching the PR's own reported count.
- View/template code, not a trust-boundary or auth path — no security-relevant behaviour change.

No blocking issues.

---

Issue #707, opened 2026-08-22
Bound DNS resolution in the outbound fetch path so a hang cannot hold the /_cron flock


svandragt:
Follow-up from #686 (opus review finding 2).

`process_feeds()` now caps itself at `set_time_limit(1800)` instead of unlimited, so a CPU-bound wedge no longer holds the cron flock forever. But on Unix `max_execution_time` does not count time blocked in a syscall, so a hung outbound operation is bounded only by its own timeout:

- The fetch itself is covered — curl sets `CURLOPT_TIMEOUT`, SimplePie gets `set_timeout(FEED_FETCH_TIMEOUT)`.
- **`resolve_validated_ip()`'s DNS lookup** (`resolve_host_ips`, which runs *before* curl connects) has no explicit ceiling. A stuck resolver could stall the run — and it holds the cron flock while it does, so every later `/_cron` returns "Already running".

## Suggested
Bound the DNS resolution step (e.g. a short timeout around `resolve_host_ips`), so no outbound step in `/_cron` can block unbounded. Low likelihood, but it is the one remaining unbounded path the flock's liveness depends on.

Files: `src/http.php` (`resolve_validated_ip`, `resolve_host_ips`).

PR #789, opened 2026-09-01
Apply DNS_RESOLVE_TIMEOUT at process boot, not just inside resolve_host_ips()

svandragt:
## Bug

`resolve_host_ips()`'s inline comment and `DNS_RESOLVE_TIMEOUT`'s own docblock (`src/http.php`) both claim every DNS lookup made through it is bounded to ~10s per nameserver, via a `putenv('RES_OPTIONS=...')` call made right before `dns_get_record()`. This bound exists specifically for issue #707: a black-holed resolver must not hold the `/_cron` flock indefinitely.

That bound is **unreliable in a persistent worker process** — php-fpm and FrankenPHP, both documented production deployment targets in `AGENTS.md`. glibc's resolver reads `RES_OPTIONS` only once, the first time *anything* in the process resolves a hostname, then caches that state. A later `putenv()` call — including `resolve_host_ips()`'s own, on every subsequent invocation in the same worker — is silently ignored.

## Why this is reachable, not theoretical

Micropub's `introspectToken()` calls `Http\fetch()` **without** the `pin` option (see `fetch()`'s own docblock: "Every other caller (Micropub's `introspectToken`) is unaffected"), so it resolves its token endpoint's hostname via plain `file_get_contents()` — with no `RES_OPTIONS` override of its own. Any Micropub POST with a bearer token is exactly this code path.

`/_cron`'s `process_feeds()` resolves one host per configured feed within a single process. In a persistent worker that has already served an earlier Micropub request (or any other un-pinned outbound call) before its next `/_cron` tick, the resolver is already locked to whatever the **system's real default** happens to be — not the documented ~10s bound — for the rest of that worker's life.

## Verified with a real, timed reproduction (no mocking)

Using a temporary `/etc/resolv.conf` pointed at a local black-holed UDP "nameserver" (never responds) and the real `src/http.php`:

- **Red (current `main`):** an unrelated `file_get_contents()` to a bad host resolves first (6.01s, the sandbox's real `timeout:2 attempts:3`). `resolve_host_ips()` (which queries A+AAAA, i.e. two lookups) then took **12.01s** (2×6.01s) — silently inheriting the earlier call's timeout, not the intended 2×10s=20s.
- **Green (this fix):** calling the new `apply_dns_resolve_timeout()` first (as `Bootstrap\bootstrap_db()` now does) made **both** the unrelated call (10.01s) and `resolve_host_ips()` (20.02s = 2×10.01s) consistently hit the intended ~10s/query bound, regardless of call order.

## Fix

Extracts the `putenv()` call into `Http\apply_dns_resolve_timeout()` and calls it as the **first line** of `Bootstrap\bootstrap_db()` — the one function every entry point (`index.php`, `bin/lamb`) calls before doing anything else network-shaped. This wins the race to be the process's first hostname resolution regardless of which code ends up needing DNS next, so every later caller in that worker's life — including `introspectToken()`'s plain fetch — inherits the correct bound. `resolve_host_ips()` keeps its own call too (harmless, and still load-bearing for a standalone script that never bootstraps).

Docblocks updated to state the actual guarantee and point at why (matching this repo's precedent for `DNS_RESOLVE_TIMEOUT` in #767 — correct the claim rather than re-architect around a platform limitation, since there's no PHP-exposed way to force glibc's resolver to re-init on demand).

## Tests

- `tests/Unit/HttpTest.php`: `testApplyDnsResolveTimeoutSetsResOptionsEnvVar` — pins the extracted function's `RES_OPTIONS` string directly.
- `tests/Unit/SessionBootstrapTest.php`: `testBootstrapDbAppliesDnsResolveTimeoutBeforeAnythingElse` — runs `bootstrap_db()` in a subprocess (it opens a real RedBeanPHP/SQLite connection via the global `R::` facade, so it can't run in the shared test process) and asserts `RES_OPTIONS` is set on return.
- `vendor/bin/codecept run Unit` — 2151 tests, 0 failures (2 skipped, unchanged from baseline).
- `vendor/bin/codecept run Acceptance` — 96 tests, 0 failures.
- `vendor/bin/phpstan analyse` — no errors. `composer lint` (`phpcs`) — clean.
- `php -l` clean on every touched file.

### DIFF
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -8,6 +8,8 @@ use Dotenv\Repository\RepositoryBuilder;
 use RuntimeException;
 use RedBeanPHP\R;
 
+use function Lamb\Http\apply_dns_resolve_timeout;
+
 /**
  * Loads the project's .env file into the process environment for the dev server.
  *
@@ -87,12 +89,20 @@ function login_password(): string
 /**
  * Initializes the database by configuring the SQLite connection and setting up the writer cache.
  *
+ * Also applies the DNS resolver timeout (see Http\apply_dns_resolve_timeout()) as
+ * the first thing any entry point does: glibc's resolver only honours a RES_OPTIONS
+ * change made *before* the process's first hostname lookup, and this runs before
+ * index.php/bin/lamb do anything else network-shaped, so it wins that race no
+ * matter which code ends up resolving a host first later in the process's life.
+ *
  * @param string $data_dir The directory path where the database file will be stored.
  * @return void
  * @throws RuntimeException If the specified directory cannot be created.
  */
 function bootstrap_db(string $data_dir): void
 {
+    apply_dns_resolve_timeout();
+
     if (!is_dir($data_dir)) {
         // 0750, not 0777: this directory holds lamb.db and the session files, so
         // only the web-server user has any business reading it. Under a permissive
--- a/src/http.php
+++ b/src/http.php
@@ -219,8 +219,36 @@ const DEFAULT_FETCH_TIMEOUT = 30;
  * resolves. Still finite either way: bounded by process_feeds()'s overall
  * 1800s /_cron time limit (see network.php), just a larger per-feed delay
  * than FEED_FETCH_TIMEOUT alone suggests when a resolver is unreachable.
+ *
+ * "Applied" only actually holds when {@see apply_dns_resolve_timeout} wins the
+ * race to be the process's first hostname resolution — see its docblock.
  */
 const DNS_RESOLVE_TIMEOUT = 5;
+
+/**
+ * Applies DNS_RESOLVE_TIMEOUT to the process via RES_OPTIONS.
+ *
+ * glibc's resolver reads RES_OPTIONS only once — the first time anything in
+ * the process resolves a hostname — then caches that state; a later putenv()
+ * call is silently ignored once that has already happened (confirmed by
+ * timing a black-holed lookup before and after a second putenv() in the same
+ * process: the second call kept the first call's timeout). So the call that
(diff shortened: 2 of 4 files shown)


svandragt:
## Automated code review

**APPROVE**

Gave this extra scrutiny as it touches DNS resolution used by the fetch/SSRF-pinning path (`src/http.php`, `src/bootstrap.php`). No merge conflicts against current `main`, CI green.

- Confirmed the extraction is behaviour-preserving: `apply_dns_resolve_timeout()` emits the exact same `RES_OPTIONS` string `resolve_host_ips()` set inline before, and `resolve_host_ips()` still calls it on every invocation (harmless no-op once already applied, load-bearing for a standalone caller that never bootstraps).
- Verified `resolve_host_ips()`'s SSRF-relevant callers (`fetch()`'s `pin` option) are untouched — this PR only changes *when* the RES_OPTIONS bound takes effect, not the hop-pinning logic itself.
- Checked out the branch and ran it directly: `vendor/bin/phpstan analyse` clean; `HttpTest`/`SessionBootstrapTest` (45 tests, 55 assertions) pass, including the subprocess-based `testBootstrapDbAppliesDnsResolveTimeoutBeforeAnythingElse`, which genuinely exercises the boot-order claim in a fresh process rather than just asserting the string. Full `vendor/bin/codecept run Unit` — 2151 tests, 3955 assertions, 0 failures (2 pre-existing skips), matching the PR's reported baseline.
- The docblock updates correctly narrow the claim (bounded *when this wins the race to be the first resolution*) rather than overstating a guarantee PHP/glibc can't actually give.

No blocking issues.

---

PR #837, opened 2026-09-13
Retry composer install past transient dist-download failures

svandragt:
## Why

PR #835 (Bump the minor-and-patch group with 2 updates) sat unmerged with auto-merge enabled because `ci` was red. Both `test (8.4)` and `test (8.5)` died seconds apart in `composer install`:

```
'/home/runner/work/lamb/lamb/vendor/composer/tmp-6965378925dfd962ec42cf50c266e509.zip' is not a zip archive.
In ZipDownloader.php line 255
Process completed with exit code 19
```

Nothing to do with the change — it's an npm-only bump. A dist mirror served a truncated body instead of a zip. Composer 2 retries transport errors but treats a bad archive at extraction as fatal, so the run just fails. A rerun went green and the PR merged itself.

## What

One composite action at `.github/actions/composer-install`, used by all six `composer install` steps across `ci.yml`, `release-verify.yml` and `release-artifacts.yml`. Three attempts, backing off 10s then 20s.

It runs `composer clear-cache` between attempts. Packagist dist zips mostly carry no `shasum`, so a truncated download can land in the cache and a plain retry would replay the same corruption.

Flag differences stay at the call sites via an `args` input; the default matches the four steps that use `--prefer-dist --no-progress`.

## Checks

- `grep -rn "composer install" .github/` — no `run: composer install` left.
- Workflows and the action parse (`yaml.safe_load`). `actionlint` isn't installed here or in devbox; CI itself is the real check.
- No CHANGELOG entry: it isn't a user-visible change.


svandragt:
Obsoleted by #838 (Install dependencies with viv instead of Composer), which removes every `composer install` from CI — there's nothing left for the retry action to wrap.

